PR #33, opened 2026-06-07
feat(search): configurable full-text search language(s)

nytafar:
Make the PostgreSQL text-search configuration for keyword search a configurable list (`FTS_CONFIGS`, default `["english"]`) instead of a hardcoded `'english'` at the three aligned index/query sites.

- Add `src/services/fts.py` as the single source of truth: `index_tsvector_sql` (concatenate `to_tsvector` over configs), `combined_tsquery` (OR `websearch_to_tsquery` over configs), and `validate_fts_configs` (startup allowlist check against `pg_ts_config`). Config names are passed as bound params with a `::regconfig` cast — never string-interpolated.
- `config.py`: `fts_configs` setting parsing JSON or CSV, normalized (strip/lowercase/dedupe), rejects empty. `NoDecode` defers env parsing to the validator so the CSV form works.
- Route the indexer tsvector writes and the search tsquery through the helpers.
- Validate configs at startup so a typo fails fast instead of producing silent zero-result searches.
- Add `rebuild_tsvectors` + `scripts/rebuild_tsvectors.py` + `make rebuild-tsvectors` to recompute the keyword index after a config change (no embeddings, no API calls).
- Default `["english"]` reproduces current behavior byte-for-byte (regression test asserts this). Setting is global, no schema migration.
- Tests (unit + DB-gated integration) and docs (README, CLAUDE.md, .env.example).

https://claude.ai/code/session_014XjSpZ3KcE59cqJQTNqF4y

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,18 @@ MCP_HOSTNAME=obsidian-mcp.example.com
 # Indexer
 INDEX_INTERVAL_SECONDS=300
 
+# Full-text (keyword) search configuration(s). PostgreSQL text-search config
+# names, applied at both index and query time. JSON list or comma-separated.
+#   english             default — English Snowball stemmer (current behavior)
+#   simple              language-agnostic, exact word forms, no stemming
+#   english,norwegian   both stemmers — good for a mixed-language vault
+#   simple,norwegian    verbatim lexemes PLUS Norwegian stems
+# A note is indexed under every listed config and a query matches if ANY
+# config's parse hits. Changing this makes stored tsvectors stale — after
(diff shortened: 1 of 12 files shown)


maxkuminov (approved):
Approving. The single-source-of-truth design in `src/services/fts.py` correctly keeps index-time and query-time configs in sync, the default `["english"]` reproduces prior behavior byte-for-byte (regression-tested), config names are bound + `::regconfig`-cast with startup validation, and the rebuild tooling is well-documented and clearly separated from `reset-embeddings`. Offline test suite passes locally (24/24); integration tests skip cleanly without a DB. Copilot review feedback was addressed in the follow-up commit.

Issue #21, opened 2026-05-30
[DISPUTED] OAuth client can self-escalate scope at /authorize because the registered client scope is never enforced


maxkuminov:
## Summary (DISPUTED finding)
OAuth client can self-escalate scope at /authorize because the registered client scope is never enforced

| | |
|---|---|
| **File** | `src/oauth/routes.py:192-196` |
| **Status** | ⚖️ Disputed — the two independent verifiers disagreed |
| **Severity (if real)** | `medium` |
| **Category** | privilege-escalation |
| **Subsystem** | auth-security |

## Description
At registration the client's requested scope is stored (OAuthClient.scope, line 138), but during /authorize the scope query/form param is validated only against the global VALID_SCOPES set via _validate_scope; it is never constrained to the scope the client registered with. A client registered with scope='read' can request scope='readwrite' at /authorize, and that value flows unmodified into OAuthCode.scope -> OAuthToken.scope. The API-key middleware maps a 'readwrite' token scope to write permission (src/mcp_server/auth.py lines 138-140), so the client gains write access it was never granted at registration. This defeats the registration-time scope restriction.

## Evidence
```python
    # Validate scope
    try:
        scope = _validate_scope(scope)   # only checks parts <= VALID_SCOPES, ignores client.scope
    except ValueError as exc:
        return JSONResponse({"error": "invalid_scope", ...}, status_code=400)
# _validate_scope: parts = set(scope.split()); invalid = parts - VALID_SCOPES; return ... or "read"
```

## Suggested fix (if confirmed)
After loading the OAuthClient in both authorize_get and authorize_post, intersect/clamp the requested scope to the client's registered scope (e.g. reject or downgrade any scope not in set(client.scope.split())) before storing it on the OAuthCode.

## Why this is disputed
Two adversarial verifiers independently re-read the code and reached **different conclusions**. This needs a human to adjudicate.

### ✅ Verifier(s) that judged it REAL
**REAL** (confidence: high, severity: medium)

[...]

PR #62, opened 2026-08-20
fix: OAuth consent screen doesn't preselect the requested scope

timkjr:
## Summary

Follow-up to #21. That issue's dispute was resolved by clamping the
*submitted* consent-form scope to what the client registered for
(`_clamp_scope` in `authorize_post`) — correctly closing the
malicious-client escalation path. But the refuting verifier's own
argument for why the original exploit was impossible pointed at a real,
separate problem that never got fixed: the consent radio buttons are
hardcoded, with zero binding to the `scope` the client actually
requested at `/authorize`.

Concretely: `authorize.html`'s "Read only" radio has `checked`
unconditionally. A client requesting `scope=readwrite` (e.g. because the
end user picked read-write access in the client's own connector
settings, as claude.ai's OAuth UI lets you do) still renders the consent
screen with "Read only" preselected. If the user clicks Approve without
noticing they need to manually re-click the other radio, they get a
read-only grant despite having asked for read-write. Reproduced this
live against a running deployment via the real claude.ai OAuth flow.

While reproducing it, found a second, related bug: the admin panel's
OAuth token list (`/admin/oauth`) shows a scope `<select>` that compares
`token.scope == 'read'` / `== 'readwrite'` by exact string equality.
`OAuthToken.scope` is a space-separated set and virtually every token
also carries `offline_access` (every DCR-registered client requests it
per `DEFAULT_CLIENT_SCOPE`), so a real scope value like `"offline_access
readwrite"` matches neither literal. Neither `<option>` gets `selected`,
so the browser defaults to the first one, "read" — silently
misrepresenting an actual readwrite grant as read-only in the admin UI.
This is what made the first bug hard to diagnose: even after getting a
correct readwrite token, the panel appeared to disagree.

## Changes

- `authorize_get`: compute `requested_write` (gated on the client
  actually being registered for `readwrite`, so an unauthorized client
  still can't get the option preselected or even shown) and pass it to
  the template; `authorize.html` preselects the matching radio instead
  of hardcoding "Read only".
- `oauth_page`: compute `has_write` per token via scope-string
  membership instead of relying on the template to do exact-equality
  comparisons against a multi-value string; `oauth.html`'s `<select>`
  now selects off that.
- `update_oauth_token_scope`: preserves the `offline_access` marker
  instead of silently dropping it when an admin flips read/readwrite via
  the panel.

None of this touches the actual security enforcement
(`_clamp_scope` in `authorize_post`, or the `readwrite in scope_parts`
permission check in `mcp_server/auth.py`) — both fixes are strictly
about the UI reflecting reality, not about what's actually granted or
enforced.

## Test plan

- [x] New regression tests: `tests/test_authorize_get_scope_preselect.py`
      (renders the real `authorize.html` via `Jinja2Templates` and
      asserts the correct radio is preselected for various
      requested-scope / client-registered-scope combinations, including
      the case where a client isn't registered for readwrite at all)
      and `tests/test_oauth_panel_scope_display.py` (renders the real
      `oauth.html` and asserts the `<select>` reflects a token's actual
      permission, plus coverage for the `offline_access`-preserving
      update endpoint).
- [x] Full existing suite passes locally (278 passed, 5 pre-existing
      skips), including `test_issue_21_registered_scope_enforced.py`
      unchanged and still green — confirms `_clamp_scope` enforcement is
      untouched.
- [x] Live-reproduced and live-verified end-to-end against a running
      deployment: disconnected/reconnected the real claude.ai OAuth
      connector requesting read-write, confirmed via direct DB
      inspection that the issued token's scope is `"offline_access
      readwrite"` post-fix (was silently downgraded to read-only
      pre-fix), and confirmed the admin panel now displays it correctly
      instead of showing "read".

### DIFF
--- a/src/control_panel/routes.py
+++ b/src/control_panel/routes.py
@@ -530,6 +530,12 @@ async def oauth_page(
                 "id": t.id,
                 "token_type": t.token_type,
                 "scope": t.scope,
+                # t.scope is a space-separated set (e.g. "offline_access
+                # readwrite"), not always exactly "read" or "readwrite" -
+                # the template needs membership, not string equality, or
+                # a readwrite token showing an offline_access marker
+                # renders as if it were read-only.
+                "has_write": "readwrite" in t.scope.split(),
                 "revoked": t.revoked,
                 "expired": False,
                 "expires_at": t.expires_at.isoformat(),
@@ -605,7 +611,11 @@ async def update_oauth_token_scope(
         return RedirectResponse("/admin/oauth", status_code=303)
     token = await _assert_oauth_token_owner(session, token_id, user)
     if not token.revoked:
-        token.scope = scope
+        # Preserve the offline_access marker (informational only - see
+        # DEFAULT_CLIENT_SCOPE comment in src/oauth/routes.py) instead of
+        # dropping it whenever an admin flips read/readwrite here.
+        has_offline = "offline_access" in token.scope.split()
+        token.scope = f"{scope} offline_access" if has_offline else scope
         await session.commit()
     return RedirectResponse("/admin/oauth", status_code=303)
 
--- a/src/control_panel/templates/authorize.html
+++ b/src/control_panel/templates/authorize.html
@@ -234,7 +234,7 @@
                 <div class="scope-label">Grant access level</div>
 
                 <label class="scope-option">
-                    <input type="radio" name="scope" value="{{ read_scope }}" checked>
+                    <input type="radio" name="scope" value="{{ read_scope }}" {% if not requested_write %}checked{% endif %}>
                     <div>
                         <div class="scope-option-title">Read only</div>
                         <div class="scope-option-desc">Search and read notes. Cannot create or edit.</div>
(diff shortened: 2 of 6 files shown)


maxkuminov:
Thanks very much for this — merged.

Both bugs were real, and the write-ups made them easy to verify. I reverted `src/` and re-ran your tests to confirm they actually fail against the old code, which they do.

It did send me down a rabbit hole. I ran a broader sweep for the same class of bug — the panel stating something about authority that the enforcement layer doesn't actually compute — and turned up a batch of related ones, so there's some follow-up tidy-up coming. They're filed as #63 through #78.

Three that came directly out of reviewing this PR:

- **#64** — the same one-row-instead-of-the-grant problem shows up in revoke, not just scope. Revoking an access token leaves the paired refresh token live, and rotation restores the grant within the hour. Fixing it with a proper grant identifier.
- **#63** — the consent preselect needs a second look. Since `/register` is unauthenticated and DCR clients register write-capable by default, binding the preselect to the requested scope means one unchanged Approve click can grant readwrite. I want to keep the intent of your fix — the user genuinely wasn't being told what the client asked for — but do it by showing the requested scope on the screen rather than preselecting it.
- **#65** — a few notes on test coverage.

None of that takes away from the contribution; the panel fix in particular was a real misrepresentation that had been there a while. Thanks again.

PR #96, opened 2026-08-22
fix(transfer): publish uploads at the umask default, not 0600 (#95)

### LINKED ISSUE #95: Transfer uploads publish at 0600; note writes publish at the umask default
## What

A file published through the transfer path (`request_upload` → `PUT /transfer/upload`, and `import_from_url`) lands with mode `0600`. Every other file in the vault — including notes written by `create_note` / `edit_note` / `write_file` — lands at the umask default (`0644` on the container).

Observed on a live upload:

```
-rw-------  1 appuser appuser  3128646 Aug 22 10:14  Attachments/chatgpt-upload-test.png   <- uploaded
-rwxrw-rw-  1 appuser appuser  2343992 Mar  4  2023  Attachments/20230304_111852.jpg       <- pre-existing
```

## Why

`vault_fs.create_temp()` opens the staging file with mode `0600`, deliberately — the docstring is right that "the content is never world-readable while it is being written". Publication is a `linkat`/`os.replace` of that same inode, so the staging mode is the *published* mode. Nothing relaxes it.

The note path already solved exactly this. `src/services/vault.py::_atomic_write_at` stages at `0600` and then calls `os.fchmod(fd, _default_file_mode())` before publishing, and `_default_file_mode()`'s docstring spells out the consequence of omitting it:

> Without that, every note the server rewrote would silently drop from the umask default (0o644 on the container) to 0o600 and become unreadable to anything else sharing the vault.

`src/services/transfer.py::stream_to_vault` drains into `vault_fs.create_temp()`'s descriptor and closes it with no such step, so the transfer path has the bug the note path documents.

## Impact

Low today, on this deployment: the container runs as `appuser` (uid 1000) and the host user shares that uid, so the vault owner can still read uploads. It bites anything that reads the bind mount as a different user or via group — another container, a backup or sync agent, a webserver. The failure is silent and per-file, so it would surface as one unreadable attachment long after the upload.

## Fix

Relax the staged descriptor to the umask default before publication, the same way `_atomic_write_at` does. `_default_file_mode()` should move to `vault_fs` (the layer both callers already depend on — `vault.py` imports `vault_fs`, not the reverse) so there is one definition rather than two.

`tests/test_vault_mutation_safety.py::test_written_note_keeps_the_umask_default_mode` is the existing gate for the note path; the transfer path needs the analogous assertion.

maxkuminov:
Closes #95.

## The bug

`vault_fs.create_temp` opens the staging file at `0600` — correctly, so the bytes are never readable by anyone else while they are being written. But publication is a `linkat`/`os.replace` of *that same inode*, so the staging mode is the published mode. Files arriving through `request_upload` / `PUT /transfer/upload` and `import_from_url` therefore landed `0600`, while every note beside them is `0644`.

Observed on a live upload:

```
-rw-------  1 appuser appuser  3128646  Attachments/chatgpt-upload-test.png   <- uploaded
-rwxrw-rw-  1 appuser appuser  2343992  Attachments/20230304_111852.jpg       <- pre-existing
```

The note path already had the fix and documents exactly this consequence — `vault._atomic_write_at` relaxes the staged descriptor with `os.fchmod` before publishing. The transfer path never got it.

## The change

- `transfer.stream_to_vault` relaxes the staged descriptor immediately after the drain, before the descriptor closes and long before publication.
- `_default_file_mode` moves from `vault` to `vault_fs` as the public `default_file_mode()`. Both writers already depend on `vault_fs` and neither depends on the other, so this is the layer that holds it — and one definition is what stops the two publish paths drifting apart again, which is how this happened.
- `create_temp`'s docstring now says its `0600` covers the staging window only and the caller must relax it before publishing.

## Scope

Deliberately not bundled: **the transfer path never `fsync`s the staged bytes.** `_atomic_write_at` does, with a comment explaining that a crash just after the rename can otherwise publish a file whose data blocks never landed — the transfer path has the same exposure. Filed separately rather than folded in here.

## Verification

- `tests/test_transfer_service.py::test_uploaded_file_keeps_the_umask_default_mode` is the new gate. Confirmed it **fails** (`0600` vs `0644`) with the `fchmod` removed, so it is a real gate and not a tautology.
- The existing note-path gate is rewired to the moved helper.
- Full suite: 1352 passed, 108 skipped.
- Postgres transfer gate against a throwaway `pgvector/pgvector:pg16`: 68 passed.
- Adversarial Codex pass on the write path: running; results posted here.

No migration, no config, no API surface change.

### DIFF
--- a/src/mcp_server/tools.py
+++ b/src/mcp_server/tools.py
@@ -2830,6 +2830,11 @@ async def import_from_url_impl(url: str, path: str, overwrite: bool = False) -> 
         return f"{e}. Nothing was written."
     except vault_fs.UnsafePath as e:
         return f"{e}. Nothing was written."
+    except vault_fs.UnsupportedFilesystem as e:
+        # Not an OSError, so the clause below does not cover it. Reachable
+        # from the publish itself and, since #95, from the staging directory
+        # refusing to be made private.
+        return f"{e}. Nothing was written."
     except OSError as e:
         return f"Could not write {rel}: {e}"
 
--- a/src/services/transfer.py
+++ b/src/services/transfer.py
@@ -1274,6 +1274,12 @@ async def _stream_locked(
                 deadline=deadline,
                 idle_timeout=idle_timeout,
             )
+            # Publication links this very inode into place, so the 0600 the
+            # staging file was created with would become the published mode.
+            # Relax it to what a plain write would have produced, exactly as
+            # `vault._atomic_write_at` does — an upload must not land less
+            # readable than the note beside it (#95).
+            os.fchmod(fd, vault_fs.default_file_mode())
         finally:
             os.close(fd)
 
--- a/src/services/transfer.py
+++ b/src/services/transfer.py
@@ -1262,9 +1262,7 @@ async def _stream_locked(
         probe_fd, name = vault_fs.open_parent(root_fd, row.path, create=True)
         os.close(probe_fd)
 
-        staging_fd = vault_fs.open_dir_beneath(
-            root_fd, vault_fs.STAGING_DIR, create=True
-        )
+        staging_fd = vault_fs.open_staging_dir(root_fd)
         fd, tmp_name = vault_fs.create_temp(staging_fd)
         try:
             size, digest, head = await _drain(
@@ -1279,6 +1277,13 @@ async def _stream_locked(
             # Relax it to what a plain write would have produced, exactly as
             # `vault._atomic_write_at` does — an upload must not land less
             # readable than the note beside it (#95).
+            #
+            # Relaxing it here rather than at publish time means the bytes are
+            # group/world-readable for as long as the gate takes, which can be
+            # minutes; `open_staging_dir` holds `.transfer-tmp` at 0700 so that
+            # window is not reachable. Keeping the descriptor open until publish
+            # would be the alternative and is worse: it pins an fd across an
+            # unbounded wait on `SELECT … FOR UPDATE`, per upload.
             os.fchmod(fd, vault_fs.default_file_mode())
         finally:
             os.close(fd)
--- a/src/services/vault.py
+++ b/src/services/vault.py
@@ -706,40 +706,6 @@ def read_file(relative_path: str, user_id: int | None = None) -> dict:
 # the name (a symlink decoy, or a leftover from a crash) cannot wedge a write.
 _TEMP_ATTEMPTS = 3
 
-# Mode a plain `open(..., "w")` produces, cached for the life of the process.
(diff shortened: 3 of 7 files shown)


maxkuminov:
### Codex adversarial review — three rounds

Convergence: **HIGH+MEDIUM → 2×HIGH+2×MEDIUM+LOW → 3×LOW**. Findings narrowed to the same class each round rather than opening new ones.

**Round 1 — FAIL.** Relaxing the staged inode before the publish gate put completed bytes at `0644` inside a `0755` `.transfer-tmp` (confirmed `0755` on the live vault) for the unbounded duration of the gate: an upload about to be *refused* was readable meanwhile, and under a group-writable umask a peer could alter it after `_drain` hashed it, so the server would publish bytes not matching the sha256 it reports. Fixed with `vault_fs.open_staging_dir` enforcing `0700` on every open.

**Round 2 — FAIL.** `open_staging_dir` trusted `fchmod` (now re-`fstat`s and refuses unless the mode is exactly `0700`, and refuses a foreign-owned directory); `prune_stale_staging` bypassed the tightening (now routed through it). Declined: reopening the staged file inside the gate to chmod it there — its own failing input is conditioned on the isolation being ineffective, which the above closes, and it costs a by-name reopen inside the publish path.

**Round 3 — FAIL, all LOW, all accepted.** The pruner's bare `os.close` in `finally` could raise onto the foreground upload path (now `close_quietly`); `import_from_url_impl` did not catch `UnsupportedFilesystem`, which is not an `OSError`, so it escaped as a raw tool exception; two negative tests did not distinguish exact-equality from "narrower than before", nor `geteuid` from `getuid`.

**Ten new tests, each confirmed to fail with its fix reverted or its check deliberately weakened.** One earlier negative check was itself invalid — an `except BaseException:` earlier in `vault_fs.py` made the code-removal a no-op — caught and redone.

**Carried out of scope, for the maintainer:** Codex asks that a fingerprint-backed overwrite preserve a restrictive incumbent's mode rather than reset it to the umask default, in `vault._atomic_write_at` as well. That changes every note overwrite and is a product decision, so it is declared and tested here rather than changed.

1364 passed, 5 skipped; Postgres transfer gate 68 passed.

maxkuminov:
**Round 4 (confirmation) — PASS**, no findings.

> No remaining or newly introduced defects found. Focused tests: 279 passed; all available non-Postgres tests: 1,364 passed. PostgreSQL gate was unavailable locally.
>
> `{verdict: PASS, findings: []}`

(The Postgres transfer gate was run separately on this branch against a throwaway `pgvector/pgvector:pg16`: 68 passed. Codex runs `--sandbox read-only` and cannot start containers.)

Issue #59, opened 2026-08-15
Migrate _atomic_write / note tools to descriptor-anchored filesystem ops (vault_fs)


maxkuminov:
Follow-up from #54's review. `validate_mutable_path` (resolve parent once, refuse symlinked leaf, act on `resolved_parent/name`) closes the static alias-retargeting vector, but `_atomic_write` still traverses the absolute pathname on each syscall (mkdir, temp create, expected read, replace). A concurrent adversary that renames the resolved parent directory and drops a symlink in its place, or repoints a symlinked vault root, between two of those syscalls can still redirect a write or a soft-delete destination.

Fix: carry an opened parent directory descriptor out of validation and do the read, exclusive temp creation, and publication relative to it (openat/renameat-style), reusing `src/services/vault_fs.py` (which already does this for the transfer routes and `delete_file`) — but allowing in-vault symlinked ancestors, which `vault_fs.open_dir_beneath` currently refuses. Also anchor `delete_note`'s `.trash` destination to the same resolved root.

Documented as an accepted limitation in CLAUDE.md until then.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #104, opened 2026-08-24
fix(vault): opt-in named-staging fallback when O_TMPFILE is unsupported

### LINKED ISSUE #103: No-clobber writes (create_note, write_file default) fail outright on TrueNAS SCALE NFS — O_TMPFILE unsupported server-side
`_create_nameless_temp()` in `src/services/vault.py` stages no-clobber writes
into an unnamed `O_TMPFILE` inode and refuses (`UnsupportedFilesystem`) when
the kernel returns `EOPNOTSUPP`/`EISDIR`/`ENOSYS`/`EINVAL`. On my production
vault mount — TrueNAS SCALE 25.10.5, NFS export over NFSv4.1 and NFSv4.2 — that
refusal fires on every `create_note` / `write_file(overwrite=False)` call.
`edit_note` / `set_frontmatter` / `write_file(overwrite=True)` (the named
staging + `renameat` path) work fine on the same mount.

**What I verified, empirically, against the live export (not guessed):**

- Raw `os.open(dir, O_WRONLY|O_TMPFILE)` fails errno 95 (`EOPNOTSUPP`), not
  `EACCES`/`EPERM` — not a permissions issue.
- Same failure as root via sudo — rules out ownership/permission checks
  entirely; the kernel rejects the operation before those would apply.
- Same failure against a second, unrelated TrueNAS export — server-wide, not
  specific to one dataset/share.
- Same failure freshly mounted under both `vers=4.1` and `vers=4.2` — not an
  NFS-minor-version gap.
- Local ext4 (same client kernel) supports `O_TMPFILE` fine, so this is not a
  general Linux client limitation — it's this NFS server implementation.
- I could not pin down *why* TrueNAS's `nfsd` rejects it (config knob vs.
  unimplemented by design) — didn't chase further, wasn't productive.
- **Update:** upgraded the NAS from TrueNAS SCALE 25.10.5 to 25.10.6 and
  retested against the live vault export. Identical failure, same errno 95
  (`ENOTSUP`). Not a version-specific gap that's already fixed upstream in
  TrueNAS itself — this is standing behavior in the current release.

**Proposal:** I've put together a patch (branch `fix/otmpfile-named-staging-fallback`
against `d575053`) that adds an **opt-in**, off-by-default fallback:

- `Settings.vault_allow_named_staging_fallback` (env
  `VAULT_ALLOW_NAMED_STAGING_FALLBACK`, default `false`) — refusal behavior is
  completely unchanged unless an operator turns this on.
- When set, `_atomic_write_at`'s no-clobber branch falls back to **named**
  staging (`_create_temp_exclusively`, the same primitive the overwrite path
  already uses successfully on this filesystem) and publishes via `link()`
  instead of `linkat` through `/proc/self/fd` — still no-clobber (`EEXIST` on
  an existing destination), but it reopens the named-staging substitution
  window `O_TMPFILE` staging exists to close (documented under "threat #59").
  That's a declared trade-off, not a silent one:
  - a `WARNING` log fires exactly once per process the first time the
    fallback is actually exercised (not merely when the flag is set),
  - `/health` gains a `vault_named_staging_fallback_active` field so it's
    visible without grepping logs,
  - the refusal error message (when the flag is off) now names the flag, so
    an operator hitting this doesn't have to go source-diving to find the
    escape valve.
- Tests added to `tests/test_anchored_note_writes.py`: default-off behavior
  unchanged, fallback still refuses to clobber, no staging litter left under
  either outcome, warns exactly once.

Happy to open this as a PR if the design's acceptable — wanted to raise it as
an issue first since it's a real architectural tension (the whole point of
`O_TMPFILE` staging is to *not* have this window) and I'd rather get your take
on the shape before sinking a PR into something you'd want done differently.

timkjr:
Closes #103.

## Summary

`_create_nameless_temp()` stages no-clobber writes (`create_note`,
`write_file(overwrite=False)`) into an unnamed `O_TMPFILE` inode and refuses
outright when the kernel rejects `O_TMPFILE`. Confirmed against a live
production mount (TrueNAS SCALE 25.10.5/25.10.6, NFSv4.1/4.2) with the write-up
in #103 — `EOPNOTSUPP` as non-root and root, reproduced on a second export,
reproduced under both NFS minor versions, absent on local ext4 on the same
client kernel. The overwrite path (named staging + `renameat`) works fine on
the same mount, so this adds an **opt-in** fallback to that proven-working
mechanism for the no-clobber path, rather than weakening the default.

- `Settings.vault_allow_named_staging_fallback` / env
  `VAULT_ALLOW_NAMED_STAGING_FALLBACK`, off by default — refusal behavior is
  unchanged unless an operator opts in.
- On, and only when `_create_nameless_temp` reports `UnsupportedFilesystem`:
  falls back to named staging (`_create_temp_exclusively`) + `link()` publish
  — still no-clobber via `EEXIST`, but it reopens the named-staging
  substitution window `O_TMPFILE` staging exists to close (documented as
  "threat #59"). Declared, not silent: a `WARNING` logs once per process on
  first actual exercise (not on flag-set), `/health` gets
  `vault_named_staging_fallback_active`, and the flag-off refusal error names
  the env var.
- Tests in `tests/test_anchored_note_writes.py`: default-off unchanged,
  fallback still refuses to clobber, no staging litter under either outcome,
  warns exactly once.

## Your three asks

1. **OpenSpec delta** — `openspec/changes/2026-08-24-named-staging-fallback/`
   (MODIFIED: Atomic write invariant in `specs/vault-write/`), stating the
   flag-off/flag-on behavior in the same register as the existing overwrite
   residual. `openspec validate --strict` wasn't runnable in my environment
   (no local CLI); the delta is hand-formatted against
   `openspec/changes/archive/2026-08-20-anchored-note-writes/` as the
   template — happy to fix anything that doesn't validate.
2. **Anchored staging** — confirmed, no code change needed: the fallback's
   `_create_temp_exclusively(dir_fd, name)` call uses `target.dir_fd`, the
   same pinned parent descriptor from `open_mutable`/`MutableTarget` that
   every other mutating write already goes through — no pathname is
   re-resolved in the fallback branch.
3. **Adversarial review** — ran one against 832fdcf before opening this PR.
   It came back clean on the core claims (dir_fd provenance, no leak on any
   exception path, overwrite path untouched, tests genuinely fail against the
   pre-fix tree) and found one real gap, fixed in the second commit:
   `_warn_named_staging_fallback_once()`'s check-then-set on the module flag
   had no lock, so two concurrent first-fallback writes from a thread-pool
   worker could both log — breaking the "exactly once per process" contract
   the comments and the new spec scenario assert. Now guarded with a
   `threading.Lock`; `/health` was already correct either way, only the
   warning cardinality was at risk.

One thing the review flagged that I did **not** change, flagging for your
call instead: `_create_nameless_temp` (pre-existing, not touched by this PR)
collapses `EOPNOTSUPP`/`EISDIR`/`ENOSYS`/`EINVAL` into the same
`UnsupportedFilesystem`, and only `EOPNOTSUPP` is what I actually reproduced
against TrueNAS. Before this PR all four led to the same hard refusal either
way; now, with the flag on, all four take the weaker fallback path. `EISDIR`/
`ENOSYS` look like legitimate documented alternate signals for missing
`O_TMPFILE` support on other kernels, but `EINVAL` is a more generic errno and
I didn't want to unilaterally narrow what the fallback accepts without your
read on whether that's warranted.

### DIFF
--- a/openspec/changes/2026-08-24-named-staging-fallback/proposal.md
+++ b/openspec/changes/2026-08-24-named-staging-fallback/proposal.md
@@ -1,0 +1,75 @@
+## Why
+
+Issue #103: `_create_nameless_temp()` stages no-clobber writes (`create_note`,
+`write_file(overwrite=False)`) into an unnamed `O_TMPFILE` inode and refuses
+with `UnsupportedFilesystem` when the kernel rejects `O_TMPFILE`
+(`EOPNOTSUPP`/`EISDIR`/`ENOSYS`/`EINVAL`). On a real production vault mount —
+TrueNAS SCALE 25.10.5/25.10.6, NFS export over NFSv4.1 and NFSv4.2 — that
+refusal fires on every no-clobber write. Verified against the live export, not
+guessed: `EOPNOTSUPP` (errno 95) as non-root and as root (rules out
+permissions), reproduced on a second unrelated export (server-wide, not one
+dataset), reproduced fresh under both NFS minor versions, and absent on local
+ext4 on the same client kernel. The overwrite path (`edit_note`,
+`set_frontmatter`, `write_file(overwrite=True)`), which stages via
+`O_CREAT|O_EXCL` named temp + `renameat` rather than `O_TMPFILE`, was tested by
+hand against the identical mount and works — so named staging + rename is a
+proven-working mechanism on this filesystem; only the unnamed-inode path is
+blocked. A vault on such a mount could not `create_note` or
+`write_file(overwrite=False)` at all.
+
+The refusal itself is deliberate — see the "threat #59" write-up in
+`src/services/vault.py` — and stays the default. This change adds an opt-in
+escape valve rather than weakening the default guarantee.
+
+## What Changes
+
+- `Settings.vault_allow_named_staging_fallback` (env
+  `VAULT_ALLOW_NAMED_STAGING_FALLBACK`), off by default. Refusal behavior is
+  completely unchanged unless an operator turns this on.
+- When set, `_atomic_write_at`'s no-clobber branch falls back to **named**
+  staging (`_create_temp_exclusively`, the same primitive the overwrite path
+  already uses successfully on this filesystem) and publishes with `link()`
+  instead of `linkat` through `/proc/self/fd` — still no-clobber (`EEXIST` on
+  an existing destination) — when `_create_nameless_temp` reports
+  `UnsupportedFilesystem`. This reopens the named-staging substitution window
+  `O_TMPFILE` staging exists to close (the parent directory holds a real,
+  observable entry between staging and publish).
+- The staged name is still created `O_CREAT|O_EXCL|O_NOFOLLOW` through the
+  same pinned parent descriptor (`MutableTarget.dir_fd`, from `open_mutable`)
+  that every other mutating write already uses — the fallback inherits the
+  anchored-write guarantees; it does not resolve any pathname the kernel
+  hasn't already been handed.
+- The trade-off is declared, not silent: a `WARNING` log fires exactly once
+  per process the first time the fallback is actually *exercised* (not merely
+  when the flag is set — `named_staging_fallback_active()` /
+  `_warn_named_staging_fallback_once()`), `/health` gains a
+  `vault_named_staging_fallback_active` boolean, and the refusal error (flag
+  off) names the env var so an operator doesn't have to source-dive to find
+  it.
+- Tests in `tests/test_anchored_note_writes.py`: default-off behavior
+  unchanged, fallback still refuses to clobber, no staging litter under
+  either outcome, warns exactly once across repeated writes.
+
+Filed as maxkuminov/obsidian-mcp#103 first (design conversation before a PR);
+accepted as shaped, with three follow-ups this change also covers: (1) this
+OpenSpec delta, (2) confirming the fallback stages through the pinned parent
+descriptor rather than reopening a pathname, (3) flagging the write-path
+adversarial review this repo runs on changes in this area.
+
+On the accepted residual: the substitution window this reopens requires an
+adversary with write access to the **destination directory** — the same
+adversary the overwrite path's residual already accepts (see "threat #59"),
+on the grounds that such an adversary could simply edit the note directly.
+This is that same declared residual, extended to the no-clobber path behind
+an explicit operator opt-in, not a new threat.
+
+## Impact
+
+- `src/config.py` — new `Settings` field.
+- `src/services/vault.py` — fallback branch in `_atomic_write_at`,
+  `_link_staged_name`, `named_staging_fallback_active`,
+  `_warn_named_staging_fallback_once`.
+- `src/main.py` — `/health` field.
+- `tests/test_anchored_note_writes.py` — new tests.
+- No database, migration, or API-surface change. No change to the overwrite
+  path or to default (flag-off) behavior.
--- a/openspec/changes/2026-08-24-named-staging-fallback/specs/vault-write/spec.md
+++ b/openspec/changes/2026-08-24-named-staging-fallback/specs/vault-write/spec.md
@@ -1,0 +1,81 @@
+## MODIFIED Requirements
+
+### Requirement: Atomic write invariant
+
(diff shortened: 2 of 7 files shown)


timkjr:
Rebased onto `main` (through 4e57a9c, the `atomic-beneath-root-writes` merge). Notes:

- Dropped this PR's own `Settings.vault_allow_named_staging_fallback` declaration and the duplicate `/health` field — both now consume what your merge introduced (`src/config.py`, `vault_fs.named_staging_fallback_active()`).
- `_create_nameless_temp`'s refusal message no longer says "until #103 lands, this write refuses whether the flag is set or not" — that sentence is now stale by construction, since this PR is what makes it land.
- Took the invitation in #105: `vault.py`'s `_staged_identity_matches` / `_require_staged_name` / `_discard_temp` now delegate to `vault_fs.staged_identity_matches` / `require_staged_name` / `discard_staged_name` instead of keeping a second copy, so the note path picks up the inode-guarded discard for free and the "consolidate the duplicated pair" follow-on you flagged is done in the same move. Also removed `vault.py`'s own copy of `named_staging_fallback_active`/the once-per-process warning in favor of calling `vault_fs.note_named_staging_exercised()` — one flag, one warning, for both paths, per D27.
- While in there: `vault_fs.note_named_staging_exercised()` had the same unlocked check-then-set race our own adversarial review caught in this PR's first draft (two concurrent first-fallback writes could both log). Added a `threading.Lock` around it — small, but it's the shared primitive both paths now call, so fixing it once covers both.
- OpenSpec delta and full test suite (1542 passed, same 5 pre-existing/unrelated failures as before this branch) both still green post-rebase.

maxkuminov:
Thanks for the PR — and for carrying the OpenSpec change with it, exactly as asked.

Heads-up per the coordination note on #103: both in-flight waves have now **merged and deployed**, so the rebase target has what you need and a bit more:

- `Settings.vault_allow_named_staging_fallback` (env `VAULT_ALLOW_NAMED_STAGING_FALLBACK`, default `false`) **already exists** in `src/config.py` — your PR should consume it rather than re-declare it.
- The `/health` field `vault_named_staging_fallback_active` exists too, fed by `vault_fs.note_named_staging_exercised()` / `named_staging_fallback_active()` — wire your once-per-process warning through the same helpers so one field covers both write paths.
- The transfer path already honours the flag with probe-selected staging modes; your note-path half completes the one-flag contract, and the base `vault-write` spec still carries the unconditional-refusal scenario your delta amends — that reconciliation is expected and welcome.
- `src/services/vault.py` gained `confirmed_publication` (a pre-publish vault-root confirmation every mutation now goes through — see the archived `durable-identity-bindings` change). Your fallback staging path will inherit it via `write_file_at`/`_atomic_write_at`; nothing extra needed, but the diff context moved a lot.
- Relevant to your fallback: **#105** — `vault._discard_temp` treats an absent staging name as a substitution and warns on every successful overwrite publish. The transfer side's `discard_staged_name` (absent ⇒ quiet, different inode ⇒ leave and warn) is the pattern; if your rebase touches that cleanup anyway, folding the #105 fix in would be welcome — otherwise we'll take it separately.

Once rebased, it goes through the same review gates our own write-path changes do (independent verifier + adversarial pass). The refusal message naming the flag on the note path already landed as a stopgap; your PR replaces the refusal with the actual fallback behind the flag.

maxkuminov:
Merged via #114 — your commits landed intact and #103 is closed. Thank you for a genuinely well-made contribution: the rebase consumed the Settings field and `/health` helpers exactly as coordinated, the fallback slots inside `_atomic_write_at` so it inherited the pre-publish root confirmation for free, and folding in the #105 consolidation was appreciated.

Full transparency on what happened at the gate, since we promised it would run: the adversarial pass found **1 BLOCKER + 1 MAJOR + 3 MINOR**, all fixed forward in the merge (`b2ee20d`) rather than bounced back — details in #114's description. The two worth knowing about:

- The BLOCKER was in an error path: `fstat` failing *after* staging reached cleanup with `staged=None`, and `discard_staged_name` unlinked the name with no identity check — a concurrent replacement could be destroyed by a no-clobber write. Now fail-closed (warn and leave the litter). For calibration: **our own transfer path had the identical pre-existing shape**, and your PR's consolidation is what put the fix in one place — so the gate finding made the whole codebase safer, not just your diff. Every first submission of our own has failed this gate too; error-path cleanup is where these always hide.
- The MAJOR was spec bookkeeping: your delta's MODIFIED requirement was copied from the base as of your rebase, and two waves had promoted new text since — archiving it verbatim would have silently erased those guarantees. Rebased onto the current text with your scenarios preserved.

Deployed to production just now; `/health` shows the fallback inactive here (ext4 supports O_TMPFILE), which is the expected reading everywhere except your NAS — where flipping `VAULT_ALLOW_NAMED_STAGING_FALLBACK=true` should now give you working `create_note`/`write_file` with the once-per-process warning naming both staging locations. If you get a chance to confirm on the TrueNAS mount, a note on #103 would be a great data point. One cosmetic residual from the verification round is filed as #115.

PR #156, opened 2026-08-28
fix(docs): double-escape surrogate examples in docstrings for Python 3.13

timkjr:
## Problem

Docstrings in `src/services/vault.py` and `src/mcp_server/tools.py` contain the literal escape `"\uD800"` (single backslash). In Python source this is an escape sequence, so the compiled docstring constant holds a real Unicode surrogate (U+D800). CPython 3.13's `compile()` refuses any module whose string constants contain surrogates — `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' ... surrogates not allowed` — so **the entire app is unimportable on Python 3.13** (`import src.services.vault` fails).

3.12 tolerates it (the surrogate sits in the docstring, harmless), which is why CI — running 3.12 per the Dockerfile — never caught this. Anyone running the server under 3.13 gets a hard import failure.

## Fix

Double the backslash (`"\\uD800"`) so the docstring constant is the literal text `\uD800` rather than a surrogate character. The rendered docstring is identical on 3.12; the module now compiles on 3.13 too.

Verified:
- double-escape compiles + renders as literal text on 3.12 and 3.13
- single-escape fails on 3.13, embeds a real surrogate on 3.12

## Files
- `src/services/vault.py` (2 docstrings)
- `src/mcp_server/tools.py` (1 docstring)

### DIFF
--- a/src/mcp_server/tools.py
+++ b/src/mcp_server/tools.py
@@ -424,7 +424,7 @@ def _response_size(result) -> int:
 def _first_unencodable_argument(bound) -> str | None:
     """The name of the first argument carrying an unpaired surrogate, or None.
 
-    A JSON-RPC argument can hold `"\ud800"`: a valid JSON string, a valid
+    A JSON-RPC argument can hold `"\\ud800"`: a valid JSON string, a valid
     Python `str`, and not a Unicode scalar value — so it cannot be encoded as
     UTF-8. Nothing good happens downstream. It cannot name a file. It cannot be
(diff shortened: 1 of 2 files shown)


maxkuminov:
Thanks @timkjr — great catch, and an exemplary PR. I reproduced it independently: the single-escape form compiles on 3.12 but fails on 3.13 exactly as you described, and with your fix every .py file in the tree compiles clean on both. Appreciate the thorough writeup and verification. Merged!

Issue #208, opened 2026-09-04
[ASVS V13.1.2] PUT /transfer/upload holds a DB connection across the semaphore wait and the whole body stream: 15 slow uploads pin the entire pool and 500 every tenant


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, requirement V13.1.2, independently verified (including an empirical QueuePool check) before filing. Severity **high**.

**Location:** `src/transfer/routes.py:324-480` (`upload_file` session scope), `:333-334` (`resolve_identity_ok` / `resolve_root_ok`), `src/services/transfer.py:1246` (semaphore acquire inside `stream_to_vault`), `:1554` (deadline checked after acquire), `src/database.py:8-9,20`

**What is wrong**

`PUT /transfer/upload` opens one `async_session()` for the whole handler, enclosing `stream_to_vault`. `claim_upload` commits, which does return the connection — but `resolve_identity_ok` and `resolve_root_ok` immediately execute again, autobegin a new transaction that is never committed, and re-check out a connection that is then held across the semaphore wait **and** the entire body stream. The semaphore is acquired inside `stream_to_vault`, strictly after that checkout, so queued uploads hold connections too, and the publish gate opens a second session, so a streaming upload pins two. `download_file` does it correctly (commit, close the session, then return the `StreamingResponse`), so upload is the outlier, not a design choice.

Pool is 5 + 10 overflow = 15 on a single worker; `pool_timeout` is the 30 s default, after which every other caller (MCP tools, OAuth `/token`, panel) gets `TimeoutError` → 500. `statement_timeout` does not help (the backends are idle in transaction) and no `idle_in_transaction_session_timeout` is set. Nothing bounds uploads below the pool: `TRANSFER_MAX_CONCURRENT_UPLOADS=4` bounds streaming only, the 10/minute limit is per IP, a one-byte-per-29-s drip holds a slot for the 600 s deadline, and the deadline is checked only after the semaphore is acquired, so a queued request's hold time is unbounded. There is no cap on pending upload tokens per key. Worst case: 15 of 15 connections pinned by one readwrite key (4 streaming + 11 idle-in-transaction in the queue). `import_from_url` does not hold a session across the fetch; it only lengthens the queue.

**Impact**

[...]

Issue #202, opened 2026-09-04
[ASVS V8.4.1] One tenant's unbounded embedding backlog starves every later tenant's indexing (no chunk cap, no round-robin, no per-tenant budget)


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, requirement V8.4.1, severity **medium** (Codex rated high; verifier downgraded because it is freshness starvation, not a request stall).

**Location:** `src/services/embeddings.py:125-149, 189-214`, `src/services/indexer.py:2313-2331, 2951-2964, 3246-3253`

**What is wrong**

`chunk_text` has no cap and `MAX_NOTE_BYTES` is 10 MiB, so one note is ~5,120 chunks; `OllamaProvider.embed_batch` sends them one sequential request at a time with only a 30 s per-call bound (the aggregate budget was deliberately removed in #127 and its absence is pinned by tests); the backlog SELECT has no `LIMIT` and drains a tenant's whole backlog before returning; `_active_user_ids()` has no `ORDER BY` and no rotation cursor, so the same tenant goes first every cycle under the global `index_pass_lock`. Re-`edit_note`ing one 10 MiB note flips `content_hash` and re-selects it every pass forever, so an admin-provisioned tenant (or an accidental bulk import) can keep every later tenant's new and edited notes out of `notes_metadata`, tsvector and embeddings indefinitely. Secondary effect: `semantic_search` query embeddings queue behind the indexer on the shared Ollama instance and degrade to slow-or-timeout. Other tenants' request paths keep serving already-indexed content; it is operator-visible only as missing `indexer_runs` rows, with no alarm.

**Recommended fix**

In order of value: (1) a max-chunks-per-note cap in `chunk_text`/`embed_note` — cheapest, bounds the worst case; (2) a persisted round-robin offset for `_active_user_ids()` (rotating a freshly re-fetched list is a no-op); (3) a per-tenant chunk/time budget checked **only at the note boundary** where `_is_paused()` already sits (`indexer.py:2349`) — mid-note preemption recreates the #127 permanent-burn defect because `embed_note` refuses partial certification. Do not split `index_pass_lock` (the #12 reindex mutual-exclusion invariant depends on it). Batching Ollama `/api/embed` inputs is a throughput win but not a fairness fix.

maxkuminov:
[...]

Issue #194, opened 2026-09-04
[ASVS V2.4.1] No rate limit or default quota on the /mcp tool surface; production keys run unlimited


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V2.4.1, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/main.py:437`

**What is wrong**

`app.mount("/mcp", APIKeyMiddleware(mcp_handler))` is the only path not covered by slowapi; the sole per-credential control is the opt-in `daily_request_limit` (src/mcp_server/tools.py:480-527), and a read-only production count (`docker exec obsidian-mcp python -c ... select count(*) filter (where daily_request_limit is not null) from api_keys where is_active`) returned `0 of 5`. There is no Traefik ratelimit middleware on the api router (container labels), no throttle on failed bearer attempts (src/mcp_server/auth.py:150-260), `semantic_search` forwards an unbounded `query` to Ollama/OpenAI (src/services/embeddings.py:176-188, 273-275; only the 61 MiB transport cap applies), and `/register` persists `oauth_clients` rows without expiry or cleanup (src/oauth/routes.py:280-296; 18 rows in production).

**Impact**

A runaway or prompt-injected agent holding a valid key can issue unlimited tool calls: bulk-exfiltrate a tenant's vault via `read_note`/`list_files`, create or delete notes at wire speed (garbage-data creation in the single source of truth), or drive embedding-provider cost with megabyte queries — with no ceiling and only after-the-fact visibility in usage_logs. A leaked key prefix can be guessed against /mcp without back-off. Unauthenticated callers can grow `oauth_clients` by ~4,300 rows/day per IP.

**Recommended fix**

Set `daily_request_limit` on every production key (the mechanism exists and is tested) and consider a non-null default for new keys; add a per-credential (and per-IP for 401s) rate limit on /mcp, either in `APIKeyMiddleware` (slowapi key on key_id/token_id) or as a Traefik `ratelimit` middleware on the api router; cap `semantic_search`/`keyword_search` query length (e.g. 8 KiB) before the embedding call; expire unused dynamically registered clients (no token ever issued after N days) via the existing token-cleanup job.

**Independent verification**

[...]

Issue #188, opened 2026-09-04
[ASVS V15.2.2] No per-credential rate or concurrency limit on /mcp tool calls


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V15.2.2, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/mcp_server/auth.py:85`

**What is wrong**

slowapi decorates only panel, OAuth and transfer routes; the /mcp mount (APIKeyMiddleware -> FastMCP) has no limiter and no semaphore. The per-key daily quota (src/mcp_server/tools.py:507-529) defaults to None (unlimited) and counts per UTC day, so a hot agent loop can issue unbounded concurrent semantic_search calls, each performing an embedding round-trip (30s timeout, embeddings.py:177) and an HNSW query (statement_timeout 60s, database.py:16) against a pool of 5+10 connections shared by all tenants and by a shared Postgres instance.

**Impact**

A single misbehaving or retry-storming agent (a real input for this product) can exhaust the DB pool and the embedding provider, degrading or timing out every other tenant's searches and writes; with statement_timeout at 60s the pool can stay saturated for a minute per wave.

**Recommended fix**

Add a per-credential token-bucket (e.g. slowapi keyed on api_key_id/oauth token id, or a small in-process limiter in _tracked) and a per-credential asyncio.Semaphore around semantic_search/find_related; make a modest default daily quota opt-out rather than opt-in; document the strategy per V15.1.3.

**Independent verification**

[...]

PR #214, opened 2026-09-05
openspec: propose mcp-rate-limits

maxkuminov:
OpenSpec proposal only — **no application code, no tests, no migration**. Design document plus six spec deltas. Revised three times against Codex pre-code review: r1 `23189e9`, r2 `5a0c827`, r3 `4f491c9` — narrowed, and **r4 `0c4b39b` — final** (see the revision comments below).

## The problem

`/mcp` is the only surface with no rate control. It is mounted outside slowapi (`src/main.py:437`), the container carries no Traefik `ratelimit` middleware, and the one per-credential control that exists — `daily_request_limit` (#162) — is opt-in (set on **0 of 5** active production keys), counts per UTC *day* so it bounds nothing instantaneous, and is exempt for OAuth by construction. Roughly half of production tool calls run on that exempt channel. A single credential can loop on `delete_note` or `read_note` at wire speed with nothing to slow it and nothing to tell it to stop.

## Why this is narrower than it started

Three rounds of cross-family review each found a **new class** of interaction between the proposed admission ceilings, the 5 + 10 connection pool and the concurrency slots — the non-convergence signal from the engineering workflow. Rather than tune another round of thresholds, **every concurrency control is deferred** to a follow-up (`mcp-concurrency-slots`) that ships in **shadow mode first**: slots computed and would-be refusals recorded, nothing enforced until real traffic shows what the ceilings would have refused. This change lands the observability half that makes that measurable. The one *proven* pool-exhaustion path was already closed by **#208**.

What remains is **rate**, which needs no arithmetic against the pool and no slot lattice.

## The control table

| L | Control | Scope key | Default | Refusal | Marker |
|---|---|---|---|---|---|
| 1 | Failed-auth budget | address slot (salted fixed table) | 60 / 300 s | **429** + `Retry-After`, pre-SQL — transport, not a tool result | none; one WARNING |
| 2 | General velocity bucket | principal | 120/min, burst 30 | in-band, sentinel line | `rate_limited` — coalesced |
| 3 | Write velocity bucket (8 mutating tools **+ `PUT /transfer/upload`**) | principal (the *minter* at redemption) | 60/min, burst 15 | in-band; **429 + `Retry-After`, claim released** at redemption | `rate_limited` / `principal_write` — coalesced |
| 4 | Vault admission (existing #66) | user | — | in-band | `no_vault_assigned` |
| 5a | Unencodable-argument (existing #149) | argument | — | in-band | `argument_not_encodable` |
| 5b | Query length cap | argument | 8,192 | in-band | `argument_too_long` — own row, not coalesced |
| 6 | Daily quota (#162), **last** pre-body gate | api key | 5,000 new keys | in-band | `over_quota` |
| 7 | Provider input rejection | argument | provider's limit | in-band `argument_too_long` **code** | `provider_input_rejected` — post-body |

Principal = `("api_key", key_id)` or `("oauth", grant_id)`.

## Load-bearing decisions

- **Gate order: buckets → vault → argument screens → quota → body.** The quota stays last, so #162's *nothing durable is consumed by a call that does not run* holds in exactly its original shape with one more gate above it.
- **The OAuth principal is the grant**, not the access token (refresh would reset it) and not `(client_id, user_id)` (two grants are independently revocable per #64). `grant_id` is NOT NULL and already on the loaded row — no extra query.
- **One caller-visible refusal shape** — a parseable `MCP-REFUSAL` final line, identical for `str` tools and structured tools via `refusal_result`; `retry_after_seconds` absent where retrying is futile. Scoped to refusals raised **inside `_tracked`**; the transport 429 is explicitly outside it.
- **The write bucket follows the bytes, not the decorator.** `PUT /transfer/upload` writes vault bytes without being a tool call, so it consumes the *minting* principal's write bucket — derivable with no extra query and no migration — and a refusal there releases the claim rather than consuming it.
- **Refusals are coalesced, not silenced**, and the coalescer entry carries a **complete immutable row template**, so a tick or shutdown flush reads no request context and needs no live credential — landing through `_log_usage`'s existing 23503 recovery. A rollover row takes the arriving refusal as its base (`suppressed = pending`); a standalone flush has none, so it carries `suppressed = pending - 1` and writes nothing when `pending` is zero. `argument_too_long` is *not* coalesced: the general bucket above it already bounds its rate.
- **A character cap cannot promise a token limit**, so the provider's own rejection is translated into the same caller-facing code under a *distinct post-body marker* — the caller-facing code and the operator-facing marker answer different questions.
- **No I/O on the common path**: dict lookups, no lock, no `await` between a bucket's read and write. Pinned by a statement-counting test.
- **Bounded state**: a salted fixed table for addresses, a capped registry with full-and-idle-only eviction and a shared overflow bucket for principals.
- **In-process, not persisted** — one worker, one copy; the durable ceiling already lives in `quota_counters`. `--workers 1` is part of the contract.
- **Traefik is the wrong instrument** for a 401 budget: a proxy cannot see the status code, and the host's static config is outside this repo. Compose labels unchanged, decision recorded there.
- **`src/database.py` is untouched** and settings read no pool constants.

## Accepted limitations (explicit)

1. **Concurrency is unbounded** until `mcp-concurrency-slots` ships.
2. OAuth grants and grandfathered NULL-limit keys have **velocity bounds only** — no durable daily ceiling (owner-approved).
3. The transport 429 sits outside the in-band refusal contract.
4. A hard kill loses at most one coalescing interval of pending refusal counts per key.
5. Past 10,000 tracked principals, shared overflow entries lose per-principal attribution but not their count.

## Non-goals

Concurrency (follow-up, above). Indexer fairness (#202) is a sibling change. Expiring unused DCR clients is an accepted residual on #194.

## Slices

A (`refusals.py`, `rate_limits.py`, `config.py`, `session.py`, `auth.py`, `tools.py`, `indexer.py`, `main.py`, `Dockerfile`) → then B (predicate + attribution doc), C (quota clock read + default-quota surface), D (`embeddings.py`), F (`transfer/routes.py`, `services/transfer.py`), E (docs, `CLAUDE.md`, `search.md`, compose comment, env/README) in parallel. No file is owned twice. Verification covers the full suite, `openspec validate --strict`, the verifier, an adversarial Codex pass **scoped away from concurrency**, deploy, and a live MCP exercise naming the tools called.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed.

## Numbers that want the owner's word

Every default is a setting. `design.md` → Open Questions lists six: the two bucket rates, `DEFAULT_DAILY_REQUEST_LIMIT` 5,000 vs 2,000, the auth-failure budget, whether the five live keys should get a limit at deploy, and whether shadow-mode-first is the right shape for the follow-up — all against ~1,600 production tool calls per 30 days.

Refs #194 #188


maxkuminov:
## Revision r1 — Codex pre-code review folded in (`23189e9`)

3 BLOCKER / 6 MAJOR / 5 MINOR / 1 NIT. Twelve folded in, three rejected with reasons (recorded under **Rejected findings** in `design.md`).

### BLOCKERs

- **Pool still exhaustible by a valid credential.** New **Layer 0**: a process-wide in-flight ceiling in `APIKeyMiddleware` (`MCP_MAX_INFLIGHT_REQUESTS`, 64) checked *before any session or query*, answering 503 + `Retry-After`. Refusal recording is bounded by **coalescing** — one row per `(principal, tool, marker)` per 10 s carrying `suppressed: <n>`, with the perf page's refusal count summing `1 + suppressed`. New integration test `tests/integration/test_mcp_pool_availability_pg.py`: a second tenant is served throughout a valid-credential flood.
- **Quota consumed by a call that then timed out on a slot.** Gate order is now buckets → vault → argument screens → **slots** → **quota** → body, with slots held across the quota gate and released in the same `finally`. #162's invariant restated as the rule it always was: *nothing durable is consumed by a call that does not run*. Conflicting scenarios in `usage-quotas` rewritten.
- **Refusals not caller-visible as typed results.** New `src/services/refusals.py` defines one shape rendered as a final sentinel line — `MCP-REFUSAL {"code":…,"scope":…,"limit":…,"limit_unit":…,"retry_after_seconds":…}` — appended to the existing prose for `str` tools and carried identically into a structured tool's error field via `refusal_result`. The three existing pre-body refusals adopt it **additively** (prose unchanged). `retry_after_seconds` is absent where retrying is futile, so no refusal invites a loop that cannot end.

### MAJORs

- **Principal → `oauth_tokens.grant_id`** (NOT NULL, indexed, migration 014, already on the row the middleware loads — no extra query). Two grants for one client+user are independently revocable and now hold independent allowances; tested across a refresh and across two grants.
- **Per-tenant embedding cap** (`MCP_MAX_CONCURRENT_EMBEDDING_PER_USER`, 1) inserted as slot level 3, so one user cannot take both embedding slots through an API key *and* an OAuth grant.
- **Write bucket** (`MCP_WRITE_RATE_LIMIT_PER_MINUTE` 60, burst 15) on the eight mutating tools, taken in addition to the general bucket. The general bucket is now described only as a **velocity** bound; design says plainly that 60/min still reaches every note here in ~45 min, that totality is bounded by the daily quota, and that `permanent=False` is the recovery path.
- **Bounded limiter state.** Hard cap + amortised TTL sweep; eviction only when **full and idle** (a depleted entry would grant free capacity; a busy one has live counts); overflow → shared overflow bucket. Addresses instead use a fixed-size **per-process-salted** hash table — bounded by construction, collisions only stricter. 100k-cardinality tests.
- **Sum relationship enforced**, not documented: boot fails unless global ≥ embedding + vector + write (12 ≥ 10 at defaults). The acquisition-order argument depends on it.
- **One textual "off"**: empty / `null` / `none`, via a single shared `BeforeValidator`, tested **through a real env file**.

### MINORs / NIT

Threshold inclusivity (refuse when count ≥ limit; a refused request does not increment) and `Retry-After` = seconds left in the window, in one helper, parameterised over **every** 401 branch; address-less requests charged to a shared reserved slot rather than exempted. Query-cap prohibition qualified — no quota or search statement, **usage row still written**. "NULL-limit keys unchanged" scoped to quota accounting only. Panel: blank submitted field = explicit unlimited, no POST-side substitution; `keys_page` passes the default to the template. Distinct markers `rate_limited` vs `slot_timeout` vs `argument_too_long`. Slot wait described as **approximately** the budget.

### Rejected (with reasons, in `design.md`)

1. **"…or the gate refusal writes no row."** Coalescing instead — a limiter invisible in the log is one nobody can diagnose or size, and `/admin/performance` is where an operator looks. Totals stay exact via `suppressed`.
2. **A per-tenant *reservation* for the embedding class.** Unimplementable at two slots and N tenants; the **cap** is folded in instead. A reservation that cannot guarantee a slot breaks under exactly the load it exists for.
3. **The write bucket as a blast-radius fix.** Accepted as a control, rejected as a *solution* — recorded so 60/min is not mistaken for a guarantee.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed. Open Questions in `design.md` now carries seven numbers for the owner, including the write-bucket 60/15.

maxkuminov:
## Revision r2 — Codex round 2 folded in (`5a0c827`)

2 BLOCKER / 7 MAJOR / 2 MINOR. Ten folded in, one rejected with a reason, four residuals now recorded as **explicit accepted limitations** (this being the last spec round).

### BLOCKERs

- **Identity-blind ceiling was neither fair nor pool-safe.** Now three pre-SQL gates: **L0a** per-credential in-flight cap (8) keyed on the SHA-256 the middleware *already computes* for its lookup — in-memory index only, never logged, never persisted, held in a per-process-salted fixed table so caller-mintable cardinality is bounded; **L0b** identity-blind ceiling lowered 64 → **32** (so ≥ 4 distinct credentials always admissible, none above a quarter); **L0c** an authentication-concurrency semaphore (3) held *only* across the middleware's own session. L0c is the gate that actually makes the claim true — 32 admitted requests could each be inside their auth session at once, demanding 32 of 15 connections, which neither L0a nor L0b bounds. **Arithmetic stated and boot-enforced:** 3 auth + 9 bodies = 12 of 15, reserving 3 for the indexer, panel and `/health`; the auth session is released before the body starts, so the two are additive not multiplicative. Cross-tenant flood test kept and extended to a single-credential flood.
- **Coalescing had no flush lifecycle.** Key is now `(principal, tool, marker, **scope**)`. Refusals inside an open window increment an in-memory pending count and issue **no statement of any kind** — explicitly no UPDATE, which is the amplification Codex flagged. Pending flushes as `suppressed` on the next row for that key, or by the next `_log_usage` on any path, the indexer tick, or shutdown. Visibility relaxed to **"within one interval under traffic, and in all cases by the next tick or shutdown"**.

### MAJORs

- **Totality claim scoped + owner residual recorded.** D8 and the `usage-quotas` requirement now state plainly that **OAuth grants and grandfathered NULL-limit keys have velocity bounds only** — no durable daily ceiling — with the reasons each closure was rejected (#162's operator-lockout argument; grandfathering). No new durable bound added, per the owner decision.
- **Global ceiling blockable by unclassified reads → fixed structurally, not by reordering.** Added an **`other` class** covering every previously unclassified tool, and the boot check now enforces `global ≥ Σ of all four` (round 1 summed three — that *was* the hole). This yields the invariant: at most Σ calls can hold a class slot, so **a call holding its class slot is guaranteed the global slot**, and the ordering question dissolves. *Rejected* Codex's "acquire global first": it makes a task hold the global slot while waiting on a class slot, so one saturated class parks tasks that block every tenant — strictly worse than the problem it fixes.
- **Per-tenant caps on `vector` and `write`** added alongside `embedding`; each is strictly below its class ceiling, enforced at boot.
- **Shared overflow bucket** acknowledged in the requirement text as the bounded-memory trade-off — one overflowing principal can refuse another, only past 10,000 tracked principals; fail-open and fail-closed both rejected with reasons.
- **Quota retry interval** now comes from `Admission`'s own decision instant (it carried only `day`); `retry_after_seconds = ceil(reset_at − decided_at)`, nothing downstream re-reads the clock. UTC-midnight boundary test added.
- **Char cap ≠ provider token limit.** Kept the 8,192 cap **and** added translation of the provider's input-limit error into the same caller-facing `argument_too_long` code with the provider's reason. Its usage marker is deliberately **different** — `provider_input_rejected`, **post-body**, *not* in the pre-body predicate — because the body ran and made a network call; enumerating it would drop a real round trip out of the percentiles. New Slice D owns `embeddings.py` so it stays file-disjoint from Slice A. Non-ASCII test added.
- **`mcp-request-routing` contradiction** resolved with a new MODIFIED delta: vault resolution is required before every *body*, qualified for a call already refused by an earlier gate, plus two scenarios pinning that a rate-refused call resolves no vault and reveals nothing, and that an admitted-by-rate unassigned caller still gets the unchanged `no_vault_assigned` refusal.

### MINORs

Per-provider worst case documented (Ollama 30 s; **OpenAI up to 60 s × 3 attempts ≈ 180 s**) with the slot-wait assumption reassessed: the budget is deliberately *not* sized to it (that is the unbounded-queue failure), so under a slow provider `semantic_search` degrades into fast typed refusals and `retry_after_seconds` is documented as a hint, not a promise — an accepted limitation with the operator levers named. 100,000-distinct-address **and** 100,000-distinct-fingerprint collision tests added to the admission-gate slice.

### Accepted limitations (explicit, per the last-round instruction)

1. OAuth grants and grandfathered NULL-limit keys: velocity bounds only, no durable daily ceiling.
2. A hard kill (SIGKILL/OOM) loses at most one coalescing interval of pending refusal counts per key — the alternative is a durable write per refusal.
3. The shared overflow bucket past 10,000 tracked principals can let one overflowing principal refuse another.
4. `slot_timeout`'s `retry_after_seconds` is a floor and a hint; a slot can be held for a full provider call.

### Numbers changed this round

Global tool ceiling 12 → **9** (= 2 embedding + 2 vector + 2 write + 3 other), driven by the pool arithmetic; in-flight ceiling 64 → **32**; new `MCP_MAX_INFLIGHT_PER_CREDENTIAL` 8, `MCP_MAX_CONCURRENT_AUTHENTICATIONS` 3, `MCP_RESERVED_CONNECTIONS` 3, `MCP_INFLIGHT_TABLE_SIZE` 4096. Open Questions now carries seven, including the pool budget as its own item.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed.

maxkuminov:
## Revision r3 — narrowed (`4f491c9`)

Round 3 was the third consecutive round to produce a **new class** of finding in the same area — pool arithmetic, then tenant-level admission via rotated tokens, then the refusal rows' own pool usage. That is the workflow's non-convergence signal, so rather than a fourth round of threshold patches the change is **narrowed**: everything that bounds *concurrency* is deferred whole; what ships is *rate*, which needs no arithmetic against the pool and no slot lattice.

### Moved to the follow-up `mcp-concurrency-slots`

- L0a/L0b/L0c pre-authentication in-flight ceilings (identity-blind, per-credential fingerprint, authentication-concurrency semaphore)
- the per-class / per-tenant / per-principal / global slot lattice and the "every tool has a class" reservation invariant
- the `slot_timeout` refusal and marker
- the pool-budget and class-sum boot validators (so `src/config.py` reads no pool constants and **`src/database.py` is untouched**)
- the two carried-forward rejected findings (global-first acquisition; per-tenant reservation)

`design.md` has a **"Deferred to `mcp-concurrency-slots`"** section stating why and committing it to **shadow mode first** — slots computed, would-be refusals recorded with a `shadow` flag, nothing enforced until real traffic shows what the ceilings would have refused. That is only possible because this change lands the observability half. It also records that **#208 already closed the one proven pool-exhaustion path** (`PUT /transfer/upload` holding a connection across a semaphore wait), and that tool-surface concurrency is unbounded until the follow-up — declared, not implied. `tasks.md` 6.9 files the issue at archive time.

### Round-3 findings folded in (they apply to the retained controls)

- **Coalescer self-containment.** The entry now stores a **complete immutable row template** — owner, `key_id`/`oauth_token_id`, the denormalised actor triple, tool, marker, scope, bounded params — captured when the window opens. A deferred flush reads **no ContextVar** and depends on **no live credential**; a flush whose FK no longer resolves lands through `_log_usage`'s existing 23503 recovery (clear the ids, keep `actor_*`), which is exactly what #77 built it for. `src/services/indexer.py` (tick flush) and `src/main.py` (shutdown flush **before `engine.dispose()`**) are added to Slice A's ownership. Cardinality is bounded by the same registry cap, overflowing into one shared entry per marker. Tested with ContextVars cleared *and* the credential deleted.
- **`argument_too_long` is not coalesced.** It sits below the general bucket, so its rate is already bounded to `MCP_RATE_LIMIT_PER_MINUTE` per principal — a second mechanism would buy nothing. Spec, design and tasks now agree; only `rate_limited` is coalesced.
- **The 429 is transport, not a tool result.** The refusal-shape requirement is narrowed to refusals raised **inside `_tracked`**, and the L1 429 is recorded as an accepted limitation with the reason: there is no tool call, no principal and no usage row, and fabricating a tool result for an unauthenticated request would be worse than an honest HTTP error.
- **Pool-equation validator dropped**, along with any config reading of pool constants. Boot validation now checks only: both halves of a bucket set or both null, the default limit's domain, and zero rejected.
- **Slice ordering fixed.** The provider input-limit **exception type is declared in `src/services/refusals.py`** (Slice A, dependency-free), so Slice A handles and tests it with a stubbed provider without waiting on Slice D, which only makes the real providers raise it.

### Final control table

| L | Control | Scope key | Default | Refusal | Marker |
|---|---|---|---|---|---|
| 1 | Failed-auth budget | address slot (salted fixed table) | 60 / 300 s | **429** + `Retry-After`, pre-SQL — transport, no sentinel line | none; one WARNING |
| 2 | General velocity bucket | principal | 120/min, burst 30 | in-band, sentinel line | `rate_limited` / `principal` — **coalesced** |
| 3 | Write velocity bucket (8 mutating tools) | principal | 60/min, burst 15 | in-band | `rate_limited` / `principal_write` — **coalesced** |
| 4 | Vault admission (existing #66) | user | — | in-band | `no_vault_assigned` |
| 5a | Unencodable-argument (existing #149) | argument | — | in-band | `argument_not_encodable` |
| 5b | Query length cap | argument | 8,192 | in-band | `argument_too_long` — **own row** |
| 6 | Daily quota (existing #162), **last** pre-body gate | api key | 5,000 new keys | in-band | `over_quota` |
| 7 | Provider input rejection | argument | provider's limit | in-band `argument_too_long` **code** | `provider_input_rejected` — **post-body** |

Principal = `("api_key", key_id)` | `("oauth", grant_id)`. Gate order: buckets → vault → argument screens → quota → body — the quota stays last, so #162's *nothing durable is consumed by a call that does not run* is restored to exactly its original shape with one more gate above it.

### Accepted limitations (explicit)

1. **Concurrency is unbounded** until `mcp-concurrency-slots` ships.
2. OAuth grants and grandfathered NULL-limit keys: **velocity bounds only**, no durable daily ceiling.
3. The transport 429 sits outside the in-band refusal contract.
4. A hard kill loses at most one coalescing interval of pending refusal counts per key.
5. Past 10,000 tracked principals, shared overflow entries lose per-principal attribution (not the count).

### Slices — re-verified disjoint and ordered

A (`refusals.py`, `rate_limits.py`, `config.py`, `session.py`, `auth.py`, `tools.py`, `indexer.py`, `main.py`, `Dockerfile`) → then **B** (`usage_stats.py` + attribution doc), **C** (quotas clock read + default-quota surface), **D** (`embeddings.py`), **E** (docs, `CLAUDE.md`, `search.md`, compose comment, env/README) all in parallel. No file is owned twice.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed. Codex round 4 should treat concurrency and pool interactions as out of scope for this change (`tasks.md` 6.5 says so); such findings belong on the new issue.

maxkuminov:
## Revision r4 — final spec round (`0c4b39b`)

All four round-4 findings folded in; nothing rejected.

**Coalescer arithmetic (MAJOR).** `pending` is now defined as the refusals since the last written row that **no row yet represents**, and the two flush paths are specified separately because they are not symmetric. The opening refusal writes its own row with `suppressed = 0` and sets `pending = 0`. A **rollover triggered by a new refusal** writes `suppressed = pending` with that arriving refusal as the row's base. A **standalone tick or shutdown flush** has no such base, so its row must stand for one of the pending refusals itself and carries `suppressed = pending − 1` — and writes **nothing at all** when `pending` is zero, since the refusal that opened the window already has a row. `Σ (1 + suppressed)` therefore equals the refusals observed on any interleaving. Four exact scenarios pin it: 1 refusal + flush → one row, `suppressed` 0, sum 1; 5 + flush → rows 0 and 3, sum 5; 5 then a 6th after the window closes → rollover `suppressed` 4, sum 6, `pending` reset; and a mixed rollover/flush run summing exactly.

**`PUT /transfer/upload` (MAJOR).** Redemption now consumes the write bucket of the principal that **minted** the token, so the write rate is not escapable by minting capabilities and redeeming them. Verified in the code first: the transfer row already carries `key_id` / `oauth_token_id` (migration 017, with `ck_transfer_tokens_one_credential` making them exclusive) and `resolve_identity_ok`'s credential lookup already loads the minting `OAuthToken`, whose `grant_id` is NOT NULL — so `("api_key", key_id)` / `("oauth", grant_id)` is derivable at redemption with **no additional query and no schema change**. A bucket refusal is **429 + `Retry-After`** and **releases the claim** rather than consuming it, deliberately mirroring #208's `QueueTimeout` → 503 path: a capability the server declined to serve is a promise still outstanding. Minting is not charged; `import_from_url` keeps consuming the bucket at its tool call and is not charged twice. This is a new `file-transfer` spec delta (five deltas → six) and a new **Slice F** owning `src/transfer/routes.py` and `src/services/transfer.py` — file-disjoint from every other slice, running in parallel after A.

**MINORs.** Overflow coalescer entries now key on `(tool, marker, scope)`, dropping only the principal, so an overflowed row still names the tool, the marker and the control that fired. The live provider-rejection exercise is conditioned on the configured provider: with `EMBEDDING_PROVIDER=openai` expect `argument_too_long` carrying the provider's reason, and with the default Ollama verify that it is *truncation* (normal result, no refusal), with the translation branch covered deterministically by Slice D's stub tests rather than live.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed.

Issue #193, opened 2026-09-04
[ASVS V16.3.4] Exceptions inside MCP tool bodies are neither logged nor recorded in usage_logs


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V16.3.4, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/mcp_server/tools.py:643`

**What is wrong**

_tracked's wrapper (tools.py:643-731) has no except branch: an exception from `await fn(...)` skips `_log_usage` entirely (probe run: a _tracked tool raising RuntimeError produced 'usage rows written: 0'). The SDK then converts it to an error result without any logger call (mcp/server/fastmcp/tools/base.py:116-117 raises ToolError; mcp/server/lowlevel/server.py:589-590 returns _make_error_result(str(e))). The production log contains 0 'Error executing tool' lines by construction, and the health page's error ring buffer (root ERROR handler) never sees them.

**Impact**

A write tool that fails halfway (disk full during edit_note, DB fault after a move) leaves no audit row and no log line; an agent hammering a tool into repeated exceptions is invisible on /admin/usage, /admin/performance and /admin/health; a systematic backend failure (Ollama down → every semantic_search raising) surfaces only to the agents, never to the operator.

**Recommended fix**

In _tracked, wrap the body call in try/except BaseException: on failure, log at ERROR with exc_info and the tool/actor/user_id, write a usage_logs row carrying a post-body marker such as `error: 'tool_exception'` and the exception class name (classified per the usage-attribution register), then re-raise. Add a regression test asserting the row and the log record exist when the body raises.

**Independent verification**

[...]

Issue #192, opened 2026-09-04
[ASVS V16.3.2] Common authorization refusals leave no distinguishable log event


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V16.3.2, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/mcp_server/tools.py:1220`

**What is wrong**

_require_write (tools.py:1212-1225) returns a refusal string; the usage_logs row written for the call carries no marker, so a read-only credential calling create_note is indistinguishable from a successful call in /admin/usage and /admin/performance. Transfer routes answer every unusable capability token with _not_found() (transfer/routes.py:96-98, 252-256, 271-275, 396-401) and log nothing server-side. Panel 403s (routes.py:713, 1055, 1067, 233-234), CSRF 403 (csrf.py:53) and the OAuth cross-user client 403 (oauth/routes.py:127-139) are unlogged.

**Impact**

A leaked read-only key being used to probe every write tool, or an attacker enumerating transfer capability tokens, is invisible to the operator: the only trace is Traefik's status-code line for /transfer and nothing at all for the MCP write refusals.

**Recommended fix**

Add a post-body usage marker (e.g. `error: 'permission_denied'`) in _require_write callers, and a WARNING log with actor/user_id. Log capability-token refusals server-side with the SHA-256 tag pattern already used in auth.py (_redacted_prefix) and the client IP, keeping the uniform 404 toward the client. Log panel/CSRF/OAuth 403s with user_id and route.

**Independent verification**

[...]

Issue #191, opened 2026-09-04
[ASVS V16.3.1] Panel password login and all OAuth grant decisions are unlogged at the application layer


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V16.3.1, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/auth/routes.py:183`

**What is wrong**

login_submit (auth/routes.py:169-201) returns 401 on failure and sets the session on success without any log call (module has no logger). src/oauth/routes.py contains no logger: /token invalid_client/invalid_grant/PKCE failures (630-666, 808-859), /authorize consent decisions, /register (208-311) and /revoke (936-1007) all leave no event. slowapi 429s are unlogged. The only records are Traefik's 4xx/5xx access-log lines (no username, no client_id) and oauth2-proxy's own SSO log for the panel prefix.

**Impact**

Credential guessing against a user's panel password (the second factor behind oauth2-proxy) or against OAuth client secrets/authorization codes cannot be reconstructed: which username, which client_id, how many attempts, which succeeded. In a multi-user deployment that issues credentials to third-party AI clients this is the primary authentication surface.

**Recommended fix**

Emit one structured INFO/WARNING record per authentication outcome: login success/failure (username, IP, reason), OAuth token issuance and refresh (client_id, user_id, scope), token endpoint refusals (error code, client_id), DCR registrations, revocations, and rate-limit hits. Hash or omit any secret material; keep the existing constant-response behaviour toward the client.

**Independent verification**

Confirmed at the application layer: `grep -c 'logger\.\|logging\.'` returns 0 for src/auth/routes.py, src/oauth/routes.py, src/oauth/grants.py and src/api/routes.py. login_submit (src/auth/routes.py:169-201) returns 401 on failure and sets the session on success with no log call; OAuth /token, /authorize, /register, /revoke emit nothing. What does exist: API-key failures are logged at src/mcp_server/auth.py:143 (extras dropped per V16.2.4); uvicorn's access log records the 401/302 status per request with a client IP but no username or client_id. Status PARTIAL because failed attempts are at least visible as status codes in the access log and Traefik 4xx log; medium stands because the OAuth token endpoint is the primary authentication surface for third-party AI clients in a multi-user deployment and no record ties a refusal or an issuance to a client_id or user.

[...]

Issue #190, opened 2026-09-04
[ASVS V16.2.4] Intended log format never applies: RichHandler fragments records and drops structured fields


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V16.2.4, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/main.py:32`

**What is wrong**

src/main.py:26 imports src.mcp_server.server, whose module-level FastMCP(...) (server.py:91) calls mcp's configure_logging (fastmcp/server.py:245), installing a RichHandler with format '%(message)s' on the root logger. main.py:32's logging.basicConfig is then a no-op (probe: root handlers ['RichHandler'] before and after; formatter '%(message)s'). Effects visible in production logs: `extra={reason,key_prefix}` on auth_failure (auth.py:143) never rendered — the line reads only 'WARNING auth_failure auth.py:143'; timestamps are local time without offset ('[09/04/26 13:14:59]'); long records are wrapped at console width across multiple stdout lines ('INFO All notes already embedded indexer.py:2337' / '(user_id=2)'); tracebacks are boxed in 26-column frames. Alloy ingests per line, so Loki holds fragments.

**Impact**

An operator investigating a credential-stuffing burst against /mcp sees N identical 'auth_failure' lines with no reason, no token tag and no user id, and cannot correlate them; Loki queries on user_id or reason are impossible; multi-line fragments cannot be parsed. This is the same failure across every `extra=`-carrying security log line in the codebase.

**Recommended fix**

After importing the MCP server, reconfigure the root logger explicitly: `logging.basicConfig(force=True, level=..., handlers=[StreamHandler()], format=...)` (or remove the RichHandler) and use a formatter that emits extras — ideally JSON lines with UTC ISO-8601 timestamps and the record's extra dict. Add a test that emits a record with extra fields through the configured root handler and asserts they appear in the output.

**Independent verification**

[...]

PR #216, opened 2026-09-05
openspec: propose security-event-logging

maxkuminov:
OpenSpec proposal only — no application code, no tests, no schema change.

`security-event-logging` closes the four ASVS V16 logging findings as one change, because they are one defect in four places: the server cannot tell an operator what happened to it.

**#190 — the format never applies.** `FastMCP`'s constructor calls the SDK's `configure_logging()` at import time, so `src/main.py`'s `basicConfig` is a no-op and every `extra=` field is dropped, timestamps are local, and records fragment across lines before Alloy ingests them. Fix: a new `src/logging_setup.py` reconfigures the root logger after the SDK import — removing and closing every handler it did not install **except** the ops-health ring buffer's — and emits one line of JSON per record with a UTC ISO-8601 `Z` timestamp. Deliberately not `basicConfig(force=True)`: that closes the ring buffer's handler too, and the health page's survival must be a property of the code rather than of an import order.

**Fields are allow-listed, never dumped.** A fixed, typed, length-bounded set of names; everything else is dropped. Identifiers may come only from a resolved database row; anything derived from a presented credential appears solely as `token_tag` (`sha:` + 8 hex, `_redacted_prefix`'s behaviour, moved to one definition). `key_prefix` is removed from the list on purpose, and a source sweep test keeps the list and the call sites from drifting.

**#191 — one record per authentication outcome**, with reason codes: panel login success/failure (the merged failure branch is split for the reason code only — the 401 stays byte-identical), logout, bootstrap, OAuth issuance/refresh/every `/token` refusal, consent granted/denied, DCR, revocation with the rowcount currently discarded, the RFC 7009 no-ops the response must conceal, the swallowed rotation `server_error`, and every 429 through one wrapper on slowapi's handler. No password, client secret, code, token or PKCE verifier is ever a field.

**#192 — refusals become distinguishable.** `_require_write` records `params.error = 'permission_denied'` and logs once **at its single definition**, so all nine call sites inherit it; transfer refusals log the redacted tag, the trusted client IP and the reason the response withholds, with the uniform 404 byte-identical; panel/CSRF/OAuth 403s log route, method and user id. Denial logging is rate-limited per event and subject so a hostile client cannot flood Loki or evict the 100-entry error buffer — and the `usage_logs` row is never suppressed. `/admin/usage` gains an outcome badge.

**#193 — `_tracked` gains `except Exception`** (never `BaseException`, so cancellation passes through), logging at ERROR with `exc_info`, writing a best-effort `usage_logs` row with `tool_exception` and the exception class, then re-raising. A failing audit write is logged and swallowed; it can never mask the original exception.

**No migration.** `usage_logs.params` is JSONB and both markers ride in it; 024 stays unclaimed.

Contents: `proposal.md`, `design.md` (one event-catalogue table with file:line for every emission site, the allow-list and redaction rules, ten decisions with rejected alternatives, risks, non-goals, three open questions), spec deltas for the new `security-event-logging` capability plus `mcp-request-routing`, `panel-performance-views`, `panel-usage-slicing`, `file-transfer` and `panel-ops-health`, and `tasks.md` split into four file-disjoint slices (A the foundation, then B/C/D in parallel) with the verification, deploy and live-MCP-exercise gates. Every Codex audit note from the four issues is folded in explicitly. `openspec validate --strict` passes for the change and for all 28 items.

Refs #190 #191 #192 #193


maxkuminov:
Pre-code review round 1 (Codex, FAIL: 2 BLOCKER / 11 MAJOR / 4 MINOR) folded in — `7c4280e`. **No finding rejected.** `design.md` ends with a finding-to-decision map and a note on the two folded in with a narrower mechanism than proposed.

**Blockers.** "Exactly one record per outcome" is restated as **one emission per decision**, with a separate section defining what reaches the sink (INFO always; WARNING/ERROR through the suppressor, which accounts for what it withholds). The six transfer refusal reasons were not derivable: `lookup_token` and `claim_upload` are each one filtered query, so expired / wrong-direction / claimed / completed / consumed all arrive as the same `None`. Rather than re-shape those queries — they are the single-use linearizability argument — a **read-only `classify_token_refusal`** diagnoses the reason *after* the decision and **only behind `should_emit`**, so an accepted request pays nothing and a flood cannot amplify it. `_load_valid` returns a typed `TransferRefusal` for the three predicates it evaluates itself; missing credential logs **no** tag.

**The rest.** The exception guard now wraps **only** `await fn(...)` — a quota-gate failure before it and a telemetry failure after a completed write can no longer be reported as `tool_exception`. `_log_usage` returns an explicit written/not-written status, and a cancellation during the audit write re-raises the tool's original exception. The suppressor keys on resolved `user_id` or the trusted client IP — never a token tag, username or client id, so rotating bearer tokens cannot mint allowances — plus a per-subject cap across events, lazy summaries, shutdown flush, fail-open. Fields declare a provenance class (resolved / submitted / derived), `oauth_token_id` stops being smuggled through `key_id`, and messages/tracebacks get their own rule (operational context stays; credential material does not) checked with high-entropy canaries. The traceback bound is head-and-tail with the exception type and final line preserved. `src/api/routes.py` joins slice D — it calls `_assert_key_owner` and carries an uncovered inline ownership 403 at `:224`. The usage query selects the three raw values and the route maps outcome precedence. The false "bounded by the quota gate" claim is gone: admission is unbounded and belongs to `mcp-rate-limits` (#188/#194), recorded as a residual with four others. Migration 024 is explicitly unclaimed.

`openspec validate --all --strict`: 28 passed, 0 failed.

maxkuminov:
Pre-code review round 2 (Codex, FAIL: 1 BLOCKER / 5 MAJOR / 3 MINOR, all against round 1's own answers) folded in — `0008968`. **No finding rejected.** `design.md`'s finding map now covers both rounds.

**Blocker — the allowance was checked twice.** Round 1 gated the transfer diagnosis on `should_emit` while `emit` suppressed independently, and the subject it proposed for a collapsed transfer refusal (the token's owner) is knowable only *after* the diagnosis that check gates. `should_emit` is deleted. There is one accounting primitive: `permit = acquire(event, subject)` charges the allowance **once**, `emit(permit, …)` consumes it with **no second check**, and `emit(event, subject=…, …)` is the convenience form. The subject must be computable before any work the permit gates, so `transfer_refused` always keys on the **trusted client address** — including on branches where a row later resolves and the record carries `user_id`; the subject is an accounting key, not an attribution. Shared-address behaviour is specified: co-located tenants share a bucket, it affects records only (never responses, audit rows or the ability to redeem), and the summary count makes the suppression visible. Residual R6.

**Majors.** An entry holding a nonzero withheld count now **emits its summary before eviction**, with a fill-past-the-bound test. Provenance moves from a global class per field to the **field name** — `username_submitted`, `client_id_submitted`, `client_name_submitted` — plus an `EVENT_FIELDS` per-event registry that `emit` enforces (strict in tests), because round 1's classes contradicted half the catalogue: successful login logs a resolved username, DCR logs client-supplied text, and the early refresh failures have no `grant_id` at all. Every field is explicitly optional. **INFO is now suppressed and accounted like every other level** (owner decision recorded): replaying a signed consent state or a logout is otherwise unbounded on routes slowapi does not cover; summaries carry exact counts, so "how many logins succeeded" stays answerable. Success records move **after their commit** — the finding named `panel_bootstrap_admin_created`, and the same defect was present at `panel_login_succeeded` (before the `last_login_at` commit) and `panel_password_reset`, so D17 states the rule once and failed-commit tests cover all three. The secret canaries gain a **captured** half — DCR `client_secret`, the `/authorize` code, both token pairs, the `omcp_` key, the minted transfer token — since no submitted canary can cover a server-generated secret; the submitted half now enumerates every real credential position including transfer redemption.

**Minors.** A value failing its declared type is **dropped, never converted** (`user_id="x"` yields no `user_id`), and the call still never raises. The claim that asyncio re-delivers a swallowed cancellation is **withdrawn** — supersession is documented as deliberate, with the consequence stated: a `wait_for` around a call that had already failed sees the tool's exception rather than `CancelledError`, and the alternative loses the tool exception entirely. `classify_token_refusal` gets a **total precedence** (unknown → direction → state → expiry → claim race) with overlapping cases tested: expired+consumed → consumed, expired+wrong-direction → wrong direction.

`openspec validate --all --strict`: 28 passed, 0 failed.

maxkuminov:
Pre-code review round 3 (Codex, FAIL: 6 MAJOR / 3 MINOR, no blocker) folded in — `5a73faf`. **No finding rejected in any round.** This is the final spec revision; implementation starts here.

The diagnosis read introduced in round 2 could raise on a path whose entire contract is a fixed 404, so it is now caught **outside** the admission decision, reports `reason=diagnosis_failed` best-effort, and returns the byte-identical refusal (injected-failure test in Slice D). The larger gap: every security or refusal event **already in the tree** went straight through `logger.*` and so bypassed the suppressor entirely — a bounded channel sitting beside an unbounded one. Design now carries a migration table naming each site, its event, level and owning slice (the ten `auth_failure` sites, `tool_refused_no_vault`, `tool_refused_over_quota`, `usage_log_credential_gone`, the two "Failed to log usage" f-strings, the structured-result and link-rewrite warnings, both quota events, the two publication refusals, the malformed-bcrypt-hash warning, five transfer sites, three panel sites) plus an explicit line against background work — indexer, embed pass, `vault_fs` housekeeping stay on the bare logger, so suppression can never hide what the health page exists to show — enforced by a test forbidding new direct calls in the four request-path modules. Alongside that: formatter-owned names, emitter control keywords and call-site fields become three disjoint sets (`error_type`'s dual case resolved — the exception wins), excluded from the subset test and the AST collection; `actor_user_id` (who acted) and `user_id` (whom the record is about) replace the ambiguous `owner_user_id` with the admin-revokes-user-B case as a test and `/revoke` documented as carrying no acting principal; the locked publish gate keeps one generic `prepublish_revalidation_failed` as an accepted limitation with its operator-visible consequence written down (Residual R9), while the earlier re-validation splits its `or` chain and keeps three distinct reasons; `users.py` is corrected to **three** `_actor_still_privileged` sites (`:320`, `:449`, `:601`) with `create_user`'s missing lock recorded as Residual R7 rather than changed here; `panel_logout` gets `_session`-suffixed names documenting that it reads the cookie and adds no lookup; `client_ip` is enumerated per event and added to the OAuth token exchange, refresh, rotation failure and cross-user refusal, with MCP tool events omitting it and saying why (Residual R8); and `client_id` leaves the generated-secret canary list, since it is a public identifier the catalogue logs on purpose. Residuals now run R1–R9. `openspec validate --all --strict`: 28 passed, 0 failed.

Issue #206, opened 2026-09-04
[ASVS V2.2.3] Embedding and tsvector rows carry no fingerprint of the generating config: a same-dimension model swap silently mixes vector spaces, and EMBEDDING_MODEL carries no reset warning


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, independently verified with measurements before filing.

**Location:** `src/models/db.py:428-448` (`NoteEmbedding`), `src/main.py:40-66` (`_check_embedding_dim`), `src/main.py:208-217` / `src/services/fts.py:75-91` (`_validate_fts_configs`), `.env.example:127,139-140`

**What is wrong**

`note_embeddings` stores no provider/model/chunk_size/overlap; `extraction_version` fingerprints only the fence cleaner. Startup catches a **dimension** change (hard exit) and checks FTS config names exist, never what stored rows were built with. Because the embed backlog selects on `embedded_content_hash != content_hash` and tsvectors rewrite only when content changes, these are not caught: a same-dimension model swap (e.g. bge-m3 → another 1024-dim model — mixed vector spaces, cosine distance meaningless, permanent), `CHUNK_SIZE`/`CHUNK_OVERLAP` changes, and an `FTS_CONFIGS` change (recall degrades). The documented control (`make reset-embeddings` / `make rebuild-tsvectors`) is written next to the chunk, overlap, provider and FTS knobs in `.env.example` and README — but **not** next to `EMBEDDING_MODEL` or `OPENAI_EMBEDDING_MODEL`, which is exactly the unguarded case.

**Impact**

Operator-only (edit `.env`, skip a step), no attacker path — but the result is the product's #2 expensive failure (silently wrong semantic search an agent acts on unseen), permanent absent a manual reset, and the project already adopted the "derived rows are fingerprinted to their generating config" principle for the cleaner in #150.

**Recommended fix**

[...]

Issue #201, opened 2026-09-04
[ASVS V8.2.2 / V16] Embedding-provider failure is counted as a successful embed pass — #160 outage detector never fires


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, surfaced by the independent verifier while checking the V8.2.2 miss. Severity **medium**.

**Location:** `src/services/embeddings.py:424-428` (`embed_note`), `src/services/indexer.py:2449-2451, 2467-2471` (`embed_vault`)

**What is wrong**

`embed_note` catches the provider exception itself and returns 0. `embed_vault` therefore never reaches its `except … record_failure` branch, and `outcome.embedded += 1` executes unconditionally. A total Ollama/OpenAI outage writes `indexer_runs` rows reading `notes_embedded = N, error = NULL` — the exact "wall of healthy passes" that `EmbedPassResult` (#160, `indexer.py:104-116`, `docs/architecture/indexing-and-embeddings.md:220-237`) was built to prevent, and with a *positive* count, which is more misleading than the zero it was designed against. `consecutive_failures` never trips, so nobody is told the provider is down and the stale-chunk window (companion issue) runs for the whole outage. The dashboard coverage gauge (`src/control_panel/routes.py:405-425`) measures presence, not currency, so it reads 100% while every vector is stale; the true pending count exists but is only served to the reset-progress poller (`routes.py:2091-2100`).

**Recommended fix**

Make provider failure a typed outcome from `embed_note` (distinct from the legitimate zero-chunk certification) so `embed_vault` stops counting it as embedded and `record_failure` fires, restoring #160's intent. Surface the pending (stale) count on the dashboard next to the coverage bar. Regression test: provider raises → run row has `error` set and `notes_embedded` excludes the note.

maxkuminov:
Implemented in #269. Embedding results distinguish provider failure from valid zero-chunk certification. Failed attempts populate pass errors and no longer increment successfully embedded notes; the dashboard displays pending/stale work beside coverage.

Regression coverage: `tests/test_issue_201_embed_outcome.py`, `tests/test_issue_201_pass_record.py`, `tests/test_issue_201_dashboard.py`.

[...]

Issue #200, opened 2026-09-04
[ASVS V8.2.2] semantic_search / find_related serve stale chunk text with no staleness signal


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, requirement V8.2.2, severity **medium** (Codex rated high; verifier downgraded because the window is bounded by the outage and deleted notes are purged within one tick).

**Location:** `src/services/embeddings.py:515-523, 571-585` (`semantic_search`), `src/mcp_server/tools.py:1824-1841, 1883, 1960-1966` (`find_related`)

**What is wrong**

Both vector paths select `NoteMetadata` with no `embedded_content_hash` predicate and return stored `chunk_text[:500]` verbatim. After an embedding-provider failure the old vectors are deliberately kept (issue #11), so during an outage an agent receives superseded note text as a current result with nothing marking it stale. The `embed_vault` docstring (`src/services/indexer.py:2266-2273`) already names this consequence ("silently wrong … indefinitely, for an agent consumer") but it is not recorded as an accepted trade-off anywhere. "Silently wrong search results" is one of this product's named expensive failures.

Not wrong (verified): the window self-heals once the provider recovers, because a failed embed never stamps `embedded_content_hash` and the backlog re-selects every tick; deleted notes cascade out within one pass.

**Recommended fix**

Annotate, do not filter. Filtering on hash equality would make every note edited in the last 0–5 minutes, and the whole vault during an outage, vanish from `semantic_search` and trip the O(n) exact fallback (`embeddings.py:530-545`, `tools.py:1918-1932`) — worse for an agent than a slightly stale hit. Instead: `semantic_search` already loads the full `NoteMetadata`, so emit `stale: true` when `embedded_content_hash IS DISTINCT FROM content_hash` (and consider suppressing the chunk preview while keeping path/title/similarity so the agent is pushed to `read_note`). For `find_related`, add the two columns to `find_related_stmt` and mark the line. Companion: #201 (provider failure counted as success).

maxkuminov:
Implemented in #269. Both semantic_search and find_related disclose stale embeddings while retaining result membership/ranking, and stale chunk previews are withheld. find_related also discloses a stale source, including empty-result paths.

Regression coverage: `tests/test_issue_200_stale_service.py`, `tests/test_issue_200_tool_render.py`.

[...]

Issue #160, opened 2026-08-29
Panel: performance views (per-tool latency percentiles, pass timings)


maxkuminov:
p50/p95/p99 per tool and slowest-request views over usage_logs.duration_ms/response_size; embedding-pass and indexer-pass duration instrumentation + display. Wave 2, after panel-light-mode (#159). OpenSpec change: panel-performance-views.

PR #217, opened 2026-09-05
openspec: propose index-integrity-hardening

maxkuminov:
Spec-only change proposal — **no application code and no tests**. `openspec validate --strict` passes (28/28 items).

## What it proposes

`index-integrity-hardening` closes the four remaining medium-severity ASVS findings on the embed path. They are one story told four times: **the embed pass reports success it did not achieve, and every consumer downstream believes the report.**

- **#201 — a provider failure is counted as a successful embed.** `embed_note` swallows the provider exception and returns `0`, the same value it returns for a note that legitimately cleaned to zero chunks and *was* certified; `embed_vault` then increments its embedded count for both. A total outage writes `notes_embedded = N, error = NULL` — the row a healthy pass writes, with a positive count, which is more misleading than the zero #160 was designed against. The proposal makes `embed_note` return a typed outcome (`EMBEDDED`, `CERTIFIED_EMPTY`, `PROVIDER_FAILED`, `PROVIDER_CARDINALITY_MISMATCH`), counts only the certifying ones, and routes the failures through `record_failure` so the run row carries `error`. `_record_index_run`'s heartbeat deliberately stays green — #160's asymmetry is preserved.
- **#200 — stale chunk text served with no signal.** Annotate, never filter: `stale: true` where `embedded_content_hash IS DISTINCT FROM content_hash`, a stale count in the header of both vector tools (rendered even when zero), and `find_related` states once when the *source* note is itself stale. Filtering was refused because on hash equality it hides every note edited in the last five minutes, the whole vault during an outage, and trips the O(n) exact fallback on every query. **The chunk preview is withheld on a stale row** and replaced by a notice naming `read_note`; path, title, tags and similarity are kept, because the scan already refreshed them — the preview is the only field that is a verbatim quotation of the note's text and the only one that is out of date.
- **#202 — one tenant starves the rest.** Three parts, in the issue's order of value: `MAX_CHUNKS_PER_NOTE` as a declared degradation in the link-cap shape (durable `notes_metadata.chunks_truncated`, one ERROR line, surfaced in vector results — and the note **is** certified, because withholding certification is #127's permanent burn arriving by a new route); a rotation cursor persisted in the database, storing a **user id** rather than an offset so it survives a membership change and a restart; and a per-tenant chunk/time budget checked **only at the note boundary** beside `_is_paused()`, never mid-note, never before the first note, and only when the pass serves more than one active scope. `index_pass_lock` is not split.
- **#206 — no fingerprint of the generating config.** One stored settings fingerprint per derived kind — canonical JSON over `provider|model|dimensions|chunk_size|chunk_overlap` for vectors, the sorted `FTS_CONFIGS` list for tsvectors — compared at startup in `_check_embedding_dim`'s shape: **fail closed** for embeddings pointing at `make reset-embeddings`, **loud warning** for FTS pointing at `make rebuild-tsvectors`. Absent is adopted, not refused. Only the repair workflows write a fingerprint afterwards, so the FTS warning cannot silence itself. Per-row fingerprints are rejected and the rejection is argued. `.env.example` and `README.md` get the reset warning at `EMBEDDING_MODEL` / `OPENAI_EMBEDDING_MODEL`.

## Shape

- One additive migration, **023**, owning `indexer_state` (key/value, with a CHECK on the key set — a mistyped key reads as *absent*, and absent means *adopt*) and `notes_metadata.chunks_truncated`, in migration 022's reconcile-or-refuse, marker-comment, no-backfill idiom.
- Spec deltas across `index-integrity`, `search-quality`, `embedding-providers`, `schema-integrity` and `panel-ops-health`.
- `tasks.md` splits the work into seven **disjoint, file-scoped** slices for parallel worktree agents (A foundation → B/C/D in parallel → E/F in parallel → G docs), with the migration in exactly one slice, plus a verification section: `make test-schema`, the full suite on the merged result, `openspec validate --strict`, `openspec-verifier`, an adversarial Codex pass, deploy + `make db-check`, and a live MCP exercise naming the tools.
- `design.md` records the rejected alternatives, the degradation-marker table, the fingerprint format and comparison rule, an enumeration of every call site touched, the thirteen named invariants each delta must preserve, and four open questions for the owner.

Refs #201 #200 #202 #206


maxkuminov:
## Pre-code review round 1 folded in (a38ac75)

Codex's pre-code pass returned **FAIL — 7 BLOCKER, 9 MAJOR, 3 MINOR**. All 19 are folded in; nothing was rejected outright, and the two findings taken at a **narrower scope** than proposed are recorded under a new **Rejected findings** section in `design.md` so the narrowing is reviewed rather than assumed.

**Blockers**
- **FTS now fails closed**, per the owner decision. The warn-and-serve path is gone, and the ERROR-vs-WARNING inconsistency (NIT) went with it. The reasoning is in D7: a stale stemmer is not thin recall — `running` stored as the lexeme `run` makes a `simple` query for `run` match a note that does not contain the word. Reverting `FTS_CONFIGS` clears the refusal with no rebuild, which is the exit that keeps it from being an outage.
- **`EmbedNoteResult` carries a bounded `EmbedNoteFailure`** (exception class, message truncated at capture, requested/received counts), and `record_failure_detail` consumes it — without this the run row read `first: None`, since `embed_note` swallows the exception.
- **`__int__` is gone.** `total_chunks += result` would have raised `TypeError`; every caller reads `result.chunks`, and §8.1 greps for survivors.
- **The FTS fingerprint is written by an outer driver** that enumerates `SELECT DISTINCT user_id FROM notes_metadata` (every scope holding rows, `NULL` included — not the active-user set), rebuilds all of them, and commits the fingerprint in the same transaction. Any failing or unpinnable scope rolls the whole thing back and is named.
- **The reset's ordering is specified *and* enforced.** Runbook inverts to deploy-then-reset (the new image refuses and stays down, so nothing is embedding); the backstop is `_embed_vault_pinned` re-reading the stored fingerprint each pass and embedding nothing when it differs from its own.
- **"A many-chunk note completes, and certifies only on full coverage" is now MODIFIED**, defining coverage over the bounded requested set, instead of being contradicted by the cap.
- **The schema gate's asserted head moves 017 → 023** as a MODIFIED requirement, with an explicit scenario that the 013/014/016/017/022 cases still run.

**Majors** — fingerprint writes abort their maintenance transaction (swallow-on-failure now explicitly limited to the rotation cursor and the run history); the reconciliation sweep feeds the same accumulator with `attempted` defined as *provider attempts*, plus a reconciliation-only-outage scenario; `find_related` states a stale source on the empty-result path; the staleness guarantee is scoped to what the scan has committed, with the pre-scan residual declared, documented in the tool docstring, and set up explicitly in the live exercise; `max_chunks_per_note` joins the embedding fingerprint; the worst-case cross-tenant delay is recorded as *the budget plus one note's embedding time* (cap × the 30 s per-call bound), accepted per the owner rather than met with a reintroduced aggregate deadline; fairness is narrowed to the embed stage with the scan and backfill declared unbudgeted; `CHUNK_OVERLAP < CHUNK_SIZE` is validated at startup; the CLI reset gets the panel's conditional HNSW creation.

**Minors** — unreadable or unknown-`v` state values fail closed and are never overwritten (the `clean_at_version` rule in a new place); the truncation ERROR is emitted only after the certifying transaction commits; every existing test that compares an `embed_note` return to an integer, asserts the semantic result key set, or fakes a `find_related` row is now assigned to a slice by name, with the one cross-slice fake dependency handled by the stub-and-swap discipline so no two slices share a file.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed. Still spec-only: no application code and no tests.

**Open for the owner** (`design.md` § Open Questions): with FTS failing closed, a scope holding `notes_metadata` rows but no readable vault would block both the rebuild and, thereafter, startup. §8.7 adds a pre-deploy check that production has no such rows.

maxkuminov:
## Pre-code review round 2 folded in (d32867a)

**1 BLOCKER, 5 MAJOR, 3 MINOR.** All folded in. Nothing rejected outright; the one alternative mechanism not taken is recorded under **Rejected findings** in `design.md`. This being the last spec round, every residual is now an explicit entry in a new **Accepted limitations** table (L1–L10) rather than prose scattered through the design.

### BLOCKER — the per-pass fingerprint re-read was not an interlock

The review is right, and the interleaving is ordinary rather than exotic:

```
old process: read fingerprint A == A, proceed
old process: get_embeddings_batch(note)        ← seconds to minutes, no lock held
reset:       wipe column, write fingerprint B, commit
old process: certify_embedded + insert vectors ← old-model vectors, stamped, under B
```

Replaced by a single **transaction-scoped advisory lock** (`pg_advisory_xact_lock` on `INDEX_GENERATION_LOCK_KEY`, a declared constant, never derived at runtime):

- taken by **every maintenance operation before it mutates** — CLI reset, panel reset, rebuild driver;
- taken by **every certifying transaction after the provider call and before certification** — precisely the window `index-integrity` already reserves, so no lock is held across a network request — with the fingerprint re-read under it and certification refused on mismatch (nothing certified, nothing inserted, nothing deleted, row left for a later pass);
- **ordering rule stated**: the advisory lock before any row or table lock, in every transaction that takes it, so it cannot close a cycle with the existing `notes_metadata` row locks;
- transaction-scoped, never session-scoped, so a crashed holder cannot strand it in a pooled connection;
- **the exclusion branch is exempt by argument**, not omission: no provider call, no vector written, and its claim ("an excluded note has been dealt with") is true under any configuration.

The per-pass read survives only as a documented cheap early exit. `tests/integration/test_issue_206_generation_lock_pg.py` drives the exact interleaving on two real connections, plus the mirror for keyword vectors, plus "no lock held during the provider call" and "a rolled-back holder releases it".

### MAJOR

- **Keyword writers interleaved with the rebuild.** The same lock now covers every `content_tsvector` writer, and the rebuild driver takes it **before it reads its first row** — otherwise an old-config incremental write lands after the rebuild recorded its fingerprint and, since keyword vectors are only rewritten on a content-hash change, that row stays on the old config for ever. Negative race scenario added.
- **`attempted` was defined two ways.** Now one increment point: **a provider call was issued**, at that call site and nowhere else. `EmbedPassResult(attempted=len(unembedded))` becomes a counter starting at zero. Every other case is *derived* from the rule rather than listed as an exception — a zero-chunk note counts into `notes_embedded` and not into `attempted` (scenario: 400 rows, 50 empty → `400` embedded, `350` attempted). Design, requirement, tasks and tests all state the same sentence.
- **The per-owner rebuild returned `0` for two things.** Now a typed `COMPLETED(rows)` / `SKIPPED_PROVENANCE_UNSETTLED` / `SKIPPED_ROOT_UNPINNABLE`. Any retained scope that is not `COMPLETED` aborts the driver and prevents the fingerprint write, named with its reason. The test asserts the provenance-skip case specifically, since that is the one that previously looked like success.
- **NULL-owned rows under multi-user mode** (owner decision): the all-scope rebuild **aborts** naming them and their count. Neither rebuilding them against `settings.vault_path` (a tenancy violation) nor excluding them from the coverage proof (they are rows `keyword_search` can return) is acceptable. §8.7 is now an explicit pre-deploy prerequisite with the query.
- **Malformed rotation cursor**: logged once and ignored, cycle starts at the first tenant, pass completes normally, never raises. Deliberately the opposite disposition from the fingerprints, and the design says why — a cursor's worst outcome is an order; a fingerprint's is a permanently wrong answer.

### MINOR

Endpoint identity (`OLLAMA_URL` as well as `OPENAI_BASE_URL`) stays out of the fingerprint per the owner decision, with **L1** recording the consequence plainly — the fingerprint records the *configuration*, not the *artifact*, so re-pulling a mutable tag like `bge-m3` mixes vector spaces undetected — and the operator rule documented at both model keys, including that no startup check will catch it. The live exercise now builds a deterministic scan-complete/embed-blocked state by making the provider unreachable for one pass, instead of racing the five-minute tick. `test_search_phase_timing.py` and `test_issue_161_result_telemetry.py` are assigned to Slice B **in full**, both their semantic and their related row fakes. The schema-gate head move is stated plainly: the literal `017` becomes `023`.

### Accepted limitations now explicit (design.md)

L1 model artifact under an unchanged name · L2 pre-scan staleness window · L3 scan and link backfill unbudgeted · L4 cross-tenant delay = budget + cap × per-call bound · L5 all-or-nothing multi-tenant rebuild, and a scope that cannot be rebuilt blocks startup after an FTS change (three recourses, one of them instant) · L6 NULL-owned rows abort under multi-user mode · L7 raising the chunk cap forces a re-embed · L8 the first startup adopts · L9 a capped note's tail is not semantically searchable · L10 `semantic_search` hydrates full vectors.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → 28 passed, 0 failed. Still spec-only.

**Action for the owner before implementation:** run `select count(*) from notes_metadata where user_id is null;` on production — it must be 0 while `MULTI_USER_MODE` is on (L6/§8.7).

maxkuminov:
## Pre-code review round 3 folded in — final spec round (84d7c93)

**1 BLOCKER + 4 MAJOR, all folded in, nothing rejected.** `openspec validate --all --strict` → 28 passed, 0 failed. Spec-only; implementation starts from this revision.

**BLOCKER — lock ordering in the incremental pass.** Correct and serious: the pass is one transaction that mutates `notes_metadata` from its first statement (changed-note upsert, id-preserving move UPDATE, prune DELETE, `note_links` delete-and-insert, the grammar-invalidation `SET embedded_content_hash = NULL`), so acquiring the generation lock at its *tsvector write* would leave the pass holding row locks while waiting for the lock and the rebuild holding the lock while waiting for those rows — a real deadlock and a direct violation of the advisory-before-row-lock rule I had just written. The rule is now stated as a property of the **transaction**, not of the statement that consumes the fingerprint: *acquire and re-validate before the first row-locking mutation*, which for the pass is the head of its transaction. The earlier mutations are enumerated as the audit's starting point rather than as a completeness claim, and `tests/integration/test_issue_206_lock_ordering_pg.py` drives the interleaving on two connections and asserts neither side is killed. Two consequences documented rather than hidden: the pass holds the lock for its whole transaction, so `make reset-embeddings` / `make rebuild-tsvectors` **wait** for an in-flight pass (correct behaviour — a reset must not land mid-pass — and the maintenance paths must not defeat it with a short `lock_timeout`), recorded as new limitation **L5b**.

**MAJOR — no interception point in the indexer.** Also correct: `get_embeddings_batch` (`embeddings.py:521`) and `certify_embedded` (`:541`) are twenty lines apart inside `embed_note`, and the caller only ever sees the result. The acquisition and fingerprint re-read move into `embed_note` itself — **Slice B, not Slice E** — and the requirement text now says the acquisition lives in the function that owns both statements. A mismatch returns a fifth outcome, `GENERATION_MISMATCH`: nothing certified or written, **not** embedded, **not** a failure (nothing went wrong with the provider), and still an *attempt*, which is the existing `attempted` rule applying unchanged rather than gaining an exception.

**MAJOR — the all-or-nothing rebuild vs. the per-user rule.** Right that it conflicted as written. Added a MODIFIED `index-integrity` delta on "The unverified ancillary passes do nothing for a user whose provenance is not settled", carving out **only** the operator-invoked, fingerprint-recording rebuild, with the three limits spelled out: it does not reach the link backfill or any periodic pass; it does not weaken the per-user gate (the skipped user still gets nothing written); and the rule's actual purpose — one tenant not blocking another's *ongoing* indexing — is untouched because this is a one-shot an operator ran. The structural reason is stated too: a fingerprint is one row asserting something about every retained row, so "all scopes rebuilt" is not a fact establishable one user at a time.

**MAJOR — the budget was debited by the wrong number.** A chunk budget debited by chunks *embedded* is never debited when the provider fails, so a tenant whose every note fails would consume every pass without reaching its bound — #202 surviving inside its own fix. `EmbedNoteResult` now carries `chunks_submitted` and `chunks_embedded` separately; the budget debits **submitted**, so a raise and a cardinality mismatch debit exactly as a success does. Tested with `EMBED_TIME_BUDGET_SECONDS_PER_USER=0` so the chunk budget alone is exercised.

**MAJOR — inactive owners.** The rebuild driver now resolves each named owner's `vault_path` **directly and read-only inside the driver** and pins it, so an inactive owner with an assigned vault and settled provenance is **rebuilt**, not skipped. It does not widen `_vault_root` or `warm_user_vault_cache` and changes nothing about which users the periodic pass serves; only an owner with no assigned path, or an unpinnable one, is `SKIPPED_ROOT_UNPINNABLE`. This narrows **L5** materially — "an inactive user's leftover rows" is no longer a blocking state unless that user is also unassigned.

The pre-deploy prerequisite is unchanged and still the one thing needed from the owner: `select count(*) from notes_metadata where user_id is null;` must be **0** on production while `MULTI_USER_MODE` is on (L6 / §8.7).

Issue #199, opened 2026-09-04
[ASVS V8.4.1] Vault-root uniqueness check misses ancestor/descendant and aliased roots (cross-tenant read/write via admin misconfiguration)


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, requirement V8.4.1, severity **medium** (Codex rated high; independent verifier downgraded because only an admin can create the overlap).

**Location:** `src/services/vault.py:266-299` (`validate_vault_root_path`), `src/control_panel/users.py:68-81` (`_check_vault_path_unique`), `src/services/vault_fs.py:432-442`

**What is wrong**

`validate_vault_root_path` rejects only empty input, a literal `..` component, a prefix outside `/vaults/`, and non-directories. `_check_vault_path_unique` is string equality over active users' `vault_path`. So `/vaults/team` (user A) and `/vaults/team/private` (user B) are both accepted, as are two symlink aliases of one directory. The FS layer confines to the root with `RESOLVE_BENEATH` but deliberately omits `RESOLVE_NO_XDEV`, and a nested tenant root *is* beneath the outer root, so A can read, overwrite and delete B's files. The indexer would also file B's notes under A's `user_id`, so search leaks the content too. There are zero tests referencing either function. The symlink case is documented as an admin-trust boundary (README "vault_path validator does not resolve symlinks"); the ancestor/descendant case is documented nowhere, and `docs/architecture/vault-tools.md` calls the `RESOLVE_NO_XDEV` omission a containment non-issue, which is true for one tenant and false once two tenants nest.

**Why medium, not high**

The only assignment path is the admin-only `edit_user_submit` (`users.py:419-427`); the dropdown offers top-level `/vaults/*` only, so the nested path must be hand-typed into the custom-path field. The same admin can already read any vault by reassignment (README troubleshooting). No untrusted actor crosses a boundary. It stays medium because the exact-duplicate rejection is a false assurance that collisions are checked, and the blast radius is maximal under the product framing.

**Recommended fix**

[...]

PR #222, opened 2026-09-05
openspec: propose vault-root-overlap-guard

maxkuminov:
OpenSpec change proposal only — no application code, no tests, no migration.

## The problem

`_check_vault_path_unique` is string equality over active users' `vault_path`,
and `validate_vault_root_path` rejects only empty input, a `..` component, a
prefix outside `/vaults/`, and a non-directory. So `/vaults/team` (user A) and
`/vaults/team/private` (user B) are both accepted, as are two symlink or
bind-mount aliases of one directory. `RESOLVE_NO_XDEV` is deliberately omitted
and B's root *is* beneath A's, so A's write tools reach B's files and the
indexer files B's notes under A's `user_id`. Admin-only to create, which is why
it is medium — but the exact-duplicate rejection is a false assurance that
collisions are checked, and the blast radius is the two failures this product
ranks highest.

## The design

**Two checks, complementary, neither implying the other**

- **Identity** — `(st_dev, st_ino)` of an opened directory descriptor per root.
  Proves the two assignments name one directory object (symlink alias,
  same-filesystem bind mount). Proves nothing about nesting; and `st_dev` is not
  a mount identity in either direction — equal does not prove one mount (hence
  `STATX_MNT_ID` in the transfer path), unequal does not prove unrelated
  directories (a filesystem mounted inside another tenant's root).
- **Containment** — component-wise realpath ancestor test in both directions.
  Proves one root's canonical pathname lies inside the other's, including
  through a symlinked component. Component-wise, not string prefix, so
  `/vaults/team-2` is not a descendant of `/vaults/team`.

**Two enforcement points**

1. Assignment time, inside the existing `_lock_admin_guard` transaction in
   `edit_user_submit` — no second lock key. Refusal names the conflicting user.
2. Every indexer pass (startup included), before the pause check — so a symlink
   that appears after the assignment, or a `vault_path` written directly in
   psql, is caught. Publishes a process-global overlap set.

**Not in `_vault_root`**, which must stay a pure cache lookup: detection needs a
query *and* filesystem I/O per call. The gate instead consults the published
set — a refuse-only frozenset membership test, no session, no statement, no
syscall — ahead of the request snapshot, which is safe precisely because a
quarantine has no direction in which staleness admits anyone.

**Disposition: index refusal *and* tool refusal for the overlapping pair only.**
Index refusal alone leaves already-indexed foreign rows queryable and leaves the
outer tenant's `edit_note` / `delete_note` / `write_file` able to clobber the
inner tenant's files — the write path never consults the indexer. Quarantine
reaches the MCP tools (via the existing "no vault" path, with a distinct
`vault_root_overlap` marker), transfer redemption, and the panel vault browser.
No rows are deleted. Unrelated tenants keep indexing and serving.

Surfaced twice — ERROR log for the ops-health ring buffer, and the affected
users' `indexer_runs.error`, because the buffer is process-lifetime while the
misconfiguration is not — plus an admin-only condition on the dashboard strip
and the health page naming both users and both roots.

`design.md` enumerates every site that assigns (5) or reads (11) a vault root,
carries an **Accepted limitations** table (L1: a bind mount of one root to a
path *inside* another tenant's tree is caught by neither check) and ten
**Owner decisions** with defaults chosen.

## Contents

`proposal.md`, `design.md`, `tasks.md`, and spec deltas for
`panel-user-administration`, `index-integrity`, `mcp-request-routing` (one
MODIFIED requirement), `file-transfer` and `panel-ops-health`.
`npx @fission-ai/openspec@1.3.1 validate --all --strict` — 29 passed, 0 failed.

Refs #199


maxkuminov:
Codex round-1 findings folded in at c5309fb. **All nine addressed, none rejected.**

| Finding | Where it landed |
| --- | --- |
| **BLOCKER** — detection only in `run_indexer_loop`; Reindex Now and the standalone rebuild bypass it | An E1–E5 entry-point table in `design.md` (lifespan, indexer startup, periodic tick, `_reindex_background`, `scripts/rebuild_tsvectors.py` — the last is a **separate process**). One `detect_and_publish()` is the only thing that computes a snapshot, and the **per-user stage skip lives in the shared pass helpers**, not in each loop, so a sixth entry point inherits the guard by routing through the helper rather than by remembering a call. New `index-integrity` requirement stated over the property ("no pass begins unchecked") with a scenario per entry point; task 3.8 tests each. |
| **BLOCKER** — async startup pass is not startup enforcement | Snapshot is **tri-state** (never published / empty / with reasons). `_vault_root` refuses every multi-user caller with a typed `VaultRootNotReady` until a snapshot exists. The lifespan runs `detect_and_publish()` **synchronously before serving**, before the indexer task is created. Publication is one immutable object swapped in with a single assignment; a later detection that raises **retains** the previous snapshot and logs ERROR, and clearing back to never-published is explicitly forbidden. New `mcp-request-routing` requirement + 4 scenarios, `index-integrity` lifecycle requirement + 4. Fail-closed is cheap because a detector failure is *not* a per-root failure — a root that cannot be opened is a per-user verdict, so the only way the routine fails is the enumeration query failing, i.e. the DB is down and the tools are unusable anyway. |
| **MAJOR** — bind-mount-inside is a permanent hole | **Check 3** added: parse `/proc/self/mountinfo`; for every entry whose mount point is strictly inside `realpath(A)`, overlap when the entry's `major:minor` equals B's root device **and** the entry's mount root is equal to / an ancestor of / a descendant of B's filesystem-relative root (computed from B's own longest-matching entry as `entry.root` joined with `realpath(B)` relative to `entry.mount_point`). **Best-effort in the refusing direction only — the `read_dir_handle` posture verbatim**: it can add an overlap, never clear one, and its absence never quarantines anyone. Reuses `vault_fs.mount_identity_available()` / `record_mount_identity_support` rather than a second notion of mount support. Residual narrowed to L1 (other namespaces — also untraversable from here), L2 (shadowing, refusing direction, admin-visible), L3 (non-Linux/`/proc` unavailable → skipped, logged once). |
| **MAJOR** — an unopenable root is not an "overlap" | Structured reasons: `overlap(peer_user_id, relation ∈ {identical, contains, contained_by, mount_graft})` vs `root_unexaminable(errno)`, which names **no peer**. Worded separately in the panel, the log line, the `indexer_runs` row **and** the `usage_logs` marker. Dedicated `index-integrity` requirement; `panel-ops-health` scenario asserts an unexaminable root is not described as an overlap. |
| **MAJOR** — the new marker must join the pre-body predicate | New `panel-performance-views` delta **MODIFYING** "Per-tool latency aggregates" to enumerate all three markers, with the reasoning that an unlisted marker is wrong in both directions at once (its `duration_ms` pollutes the percentiles *and* the refusal is never counted). `src/services/usage_stats.py` added to slice 4's ownership; task 4.3 updates `PRE_BODY_REFUSAL_ERROR_MARKERS` + the docstring enumeration, task 4.5 adds predicate tests incl. an enumeration-vs-bind-set consistency test. |
| **MAJOR** — the users list must show a quarantined state | `mcp-request-routing`'s existing "The users list MUST NOT report a note count the tools will not serve" **MODIFIED** to cover a quarantined account, with its own reason wording, plus a scenario. Slice 2 owns `users.py` **and** `users.html`. |
| **MAJOR** — slices not disjoint | An exhaustive **owned-file map** table heads `tasks.md`. 1.6 moved into slice 2 (which owns `users.py` outright, including subsuming `_check_vault_path_unique`); `src/auth/routes.py` declared in slice 2; `usage_stats.py` in slice 4; `main.py` + `scripts/rebuild_tsvectors.py` in slice 3; `routes.py` wholly in slice 6, which therefore also owns the E4 `_reindex_background` call site (6.6). |
| **MINOR** — canonical assignment strings | The predicate takes them alongside the descriptors and returns the relation, so an equal pair keeps today's duplicate wording instead of being described as a containment. Pinned as a spec scenario. |
| **MINOR** — a paused iteration must still record | Folded into the `index-integrity` recording requirement: a paused iteration still publishes, still logs at ERROR, and still writes the per-user run rows before returning; the pause suppresses index/embed work only. Task 3.7 + scenario. Row cadence is unchanged — a running deployment already writes one row per user per tick.  |

Also added: owner decisions now number 13 (new: one orchestration; fail-closed-until-first-snapshot; check 3's best-effort posture; the quarantined users-list state), accepted limitations L1–L9, a `## Review history` table in `design.md`, a seam-check task (8.5: grep for a production caller at each of E1–E5), and round-2 Codex attack vectors including the mountinfo parser against octal-escaped path fields (`\040`, `\011`, `\134`), optional fields terminated by `-`, and shadowed entries.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → **29 passed, 0 failed**.

maxkuminov:
Codex round-2 findings folded in at 90e53a5. **All four addressed, none rejected.**

| Finding | Where it landed |
| --- | --- |
| **BLOCKER** — E3/E4 call `detect_and_publish` before the pass lock, so a stalled older detection can publish an empty result over a newer quarantine and re-admit both tenants | New design section *"One detection at a time, and a publication that cannot go backwards"* + new `index-integrity` requirement **"Detection SHALL be serialized, and a publication SHALL NOT replace a newer one"**. **One process-global `asyncio.Lock` around observation, the checks AND the publication** — the spec says holding it across the publication alone is insufficient and MUST NOT be substituted, because that permits exactly the interleaving. **Plus a monotonic sequence taken under the lock**, with the reason the two are not redundant written out: the lock is the mechanism, the sequence is the invariant, and the invariant is what survives a future caller that publishes without taking the lock. Three scenarios incl. the concurrent stale-overwrite one; tasks 1.10, 1.11, and a test in 1.16 that asserts it both with the lock in place and by publishing out-of-order directly. Noted that E5 (separate process, own loop) has its own lock and consumes only what it published. |
| **BLOCKER** — mountinfo availability must not reuse `vault_fs.mount_identity_available()` | Reversed. `vault_overlap` keeps its own `_mountinfo_available: bool \| None`, probed on first use, logged once; check 3 is attempted **independently of the transfer probe's verdict**; `record_mount_identity_support` is not called and `transfer_mount_check_available`'s semantics are untouched. The independence is written out in both directions — a 5.6/5.7 kernel (this server's `openat2` floor) has mountinfo and no `STATX_MNT_ID`, and a `/proc`-less sandbox can have the reverse — with the two failure modes named: reusing the flag would **disable check 3 on kernels that can perform it**, and would let a mountinfo failure **masquerade as a transfer-write outage on `/health`**. Reduced graft coverage gets its own statement on the health page (task 6.2, `panel-ops-health` scenario) explicitly separate from `/health`'s transfer report. Updated in the check-3 table row, L3, owner decision 8, task 1.4/1.5, and the `panel-user-administration` + `index-integrity` wording. |
| **MAJOR** — gate the assignment check on the *resulting* state | Step 0 of enforcement point 1: the condition is `normalized and new_active`, evaluated after the handler has already computed both. Requirement text updated with the reasoning — an inactive or unassigned account is outside the set the check compares against and can create no overlap, so refusing such an edit protects nothing **and removes the operator's remedy**: deactivating or unassigning is how a quarantined overlap is resolved from the panel, and a check that refuses it leaves the condition with no exit through the interface that reports it. Three new scenarios: deactivation succeeds, clearing the assignment succeeds, **reactivation is refused while the overlap stands**. Owner decision 13; tasks 2.1/2.2, tests in 2.8. |
| **MINOR** — the snapshot must carry the observed names and assignments | Reason payload is now `Overlap(peer_user_id, peer_username, peer_assignment, relation)`, and every snapshot entry additionally carries the subject's username and canonical assignment plus the **detection timestamp**, immutable. Surfaces render those and **re-read no `users` row**, labelled *"as at last check"*. Reasoning recorded: the operator's first response to the condition is to edit or delete one of the two accounts, and a render-time resolution then shows a changed path — or a blank where the deleted peer was — beside a condition still in force; recording the facts also makes the staleness honest rather than presenting stale names as current state. Scenarios in `index-integrity` and `panel-ops-health` cover correction **and peer deletion before redetection**; the "panel opens no directories" scenario now also forbids re-reading `users`. Owner decision 14; test in 6.8. |

Also: owner decisions now number 16, a round-2 row added to the `## Review history` table, slice-1/2/6 tasks renumbered so the inserts are sequential, and round-3 Codex attack vectors extended with the detection lock's scope (does anything observe or publish outside it?), whether the sequence guard actually drops an out-of-order publication, and whether the resulting-state gate leaves any path that can *create* an overlap unchecked.

`npx @fission-ai/openspec@1.3.1 validate --all --strict` → **29 passed, 0 failed**.

maxkuminov:
Codex round-3 findings folded in at aa1e250 — **final spec round**. All five addressed; the two blockers are resolved by **narrowing, not by widening the check again**.

Three rounds each produced a *new* mount configuration a mount-topology check had to grow to cover (round 1: the bind-mount graft itself; round 3: a **nested** mount of the peer's vault, which defeats the device-and-mount-root comparison, and an accessible alias of an **unexaminable** root, which is never related to anything). That is the non-convergence signal from the engineering rules: rounds producing new *classes* of finding mean the design is wrong, not that the reviewer is thorough. A mount-table check that needs widening once per round is a heuristic pretending to be a rule, and this one would have sat on the admission gate for two live tenants.

| Finding | Resolution |
| --- | --- |
| **BLOCKER** — a nested mount of B grafted into A defeats the device comparison | **Check 3 removed from this change**, together with its availability state, its health-page statement, and its parser tasks and tests. Checks 1 and 2 stay and close #199 exactly as filed. Recorded as **L1** with the destructive consequence written out verbatim; owner decision 8 is now **PENDING** (accept the residual, or fund the only sound in-namespace answer — a per-pair, per-pass tree walk — as its own change); task 8.11 files `vault-root-mount-graft-detection` at archive time whichever way it goes. |
| **BLOCKER** — an accessible alias of an unexaminable root stays served | Kept "quarantine only that user" and recorded the consequence as **L2**, same class, same follow-up. The narrower alternative (quarantine every peer whenever any root is unexaminable) stays rejected under owner decision 5: one broken mount must not take the deployment offline. |
| **MAJOR** — a slow filesystem stalls the synchronous startup detection | Observation runs in `asyncio.to_thread` under `VAULT_ROOT_OBSERVE_TIMEOUT_SECONDS` (new setting, default 10). Expiry is a per-user `root_unexaminable(timeout)` verdict — distinguishable from an `errno`, because a hung mount and a deleted directory are different incidents — and the detection continues and **publishes**; treating it as a detection failure would retain a stale snapshot on every tick a slow mount was slow. New `index-integrity` requirement + 4 scenarios. The uncancellable thread is stated, not hidden: the deadline abandons the *wait*, not the syscall (**L4**). |
| **MAJOR** — the owned-file map omits existing tests | A second ownership table assigns all 21 existing tests that resolve a vault root to the slice whose production file they exercise, incl. `test_issue_88_root_confirmed_before_publish.py` (S1), `test_issue_160_indexer_runs.py` (S3), `test_issue_91_users_list_not_served.py` (S2). `tests/conftest.py` and `src/config.py` go to slice 1; task 1.13a specifies an **autouse fixture publishing an empty snapshot** (otherwise the readiness refusal turns every multi-user test into a vault-unavailable failure), which detector-entry tests **opt out of** explicitly rather than depending on import order, plus a helper to publish a synthetic snapshot so slices 2–6 drive quarantined paths without running a detection. |
| **MINOR** — mountinfo parser hardening | Moot; deleted with the parser. |

**Final accepted-limitation list (9).** L1 a bind mount grafting a peer's vault — or a mount nested inside it — to a path inside another tenant's root is undetected: A can read, overwrite and delete every note in B's vault and A's index files B's notes under A's `user_id`, with both quarantine mechanisms silent (**owner decision pending, follow-up filed**). L2 a peer whose root aliases an unexaminable assignment cannot be related to it and stays served (same class, same follow-up). L3 both checks are point-in-time; the window is one index interval. L4 the observation deadline abandons the wait, not the syscall, so a pathological mount accumulates one parked thread per detection. L5 the guard refuses the configuration but does not un-index what a previous configuration indexed. L6 the snapshot is process-global. L7 a direct `UPDATE users SET vault_path` in psql bypasses the assignment-time check. L8 `root_unexaminable` refuses the database-backed tools too. L9 the `/vaults/*` dropdown is unfiltered.

Owner decisions now number 17, one of them (8) explicitly pending. The round-4 Codex brief marks mount-graft detection **out of scope, do not re-litigate**, and adds the observation deadline's descriptor and partial-publish paths as new vectors. `npx @fission-ai/openspec@1.3.1 validate --all --strict` → **29 passed, 0 failed**.

Issue #198, opened 2026-09-04
[ASVS V7.4.1] Panel logout does not invalidate the signed session cookie server-side


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V7.4.1, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/auth/routes.py:295`

**What is wrong**

logout() only calls request.session.clear(), which makes Starlette emit an expiring Set-Cookie; the previously issued cookie remains a valid credential until its itsdangerous timestamp exceeds session_max_age (7 days). The only server-side invalidator, users.session_version (src/auth/session.py:151), is incremented solely by admin password reset (src/control_panel/users.py:610). Demonstrated against the container's installed Starlette 1.6.0 with an in-memory app: after logout the client sees user_id None, while a replay of the pre-logout cookie returns user_id 7.

**Impact**

A panel cookie captured before logout (shared machine, proxy log, XSS on the no-CSP panel) keeps working for up to 7 days after the user believes they signed out; in multi-user production that session can mint readwrite omcp_ API keys and approve OAuth grants against that user's vault, i.e. durable agent write access. Exploitation on production also requires the oauth2-proxy cookie from the same theft, which limits but does not remove the exposure.

**Recommended fix**

Give every login a server-side identifier: either store a per-session random id in the cookie and a `user_sessions` table (id, user_id, created_at, last_seen_at, revoked_at) checked in get_active_session_user, or at minimum add `users.sessions_invalid_before` and have logout stamp the cookie's issue time so any cookie issued earlier is refused. Add a regression test that replays the pre-logout cookie and expects 302 to login.

**Independent verification**

[...]

Issue #197, opened 2026-09-04
[ASVS V6.2.2] No self-service password change for users


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V6.2.2, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/control_panel/users.py:579`

**What is wrong**

The only routes that set a password are admin-only (router.dependencies.append(Depends(require_admin_panel)) at users.py:55; reset_password at :579-613). No route in src/auth/routes.py or src/control_panel/routes.py lets a signed-in user change their own password; README.md:859 confirms 'Password reset is admin-driven only'.

**Impact**

In production multi-user mode a non-admin user who suspects their password is compromised cannot rotate it and must ask an administrator, who then learns the replacement (see V6.4.1). A compromised password plus the Google account gives write access to the user's vault and the ability to mint agent credentials.

**Recommended fix**

Add a self-service POST /admin/account/password route requiring current_password + new_password, hashing via hash_password, bumping session_version, and re-issuing the session. Keep the admin reset as a recovery path only.

**Independent verification**

Confirmed. src/auth/routes.py exposes only login (:155-200), logout (:203-206) and bootstrap register (:212-345); src/control_panel/routes.py has no password route; the only writer of password_hash outside bootstrap is src/control_panel/users.py:609 inside reset_password, on a router whose every route requires require_admin_panel (users.py:55). README.md:859 states 'Password reset is admin-driven only'. Production runs MULTI_USER_MODE (env var present on the container) with several users, so a non-admin who suspects compromise cannot rotate. L1 requirement, real gap; medium stands.

_Filed from the ASVS assessment; each issue is one finding. Fixes go through the usual OpenSpec + verifier + Codex flow._

maxkuminov:
Implemented in #269. Signed-in users can change their own password using their current password and a confirmed new password. The transaction updates the password/session version, revokes sibling sessions and retains a newly issued session for the changing user.

Regression coverage: `tests/test_issue_197_password_change.py`, `tests/test_issue_198_session_revocation.py`.

[...]

Issue #183, opened 2026-09-04
[ASVS V10.4.7] Consent page does not identify or warn about dynamically-registered clients


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V10.4.7, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/control_panel/templates/authorize.html:251`

**What is wrong**

/register is unauthenticated (routes.py:206) and every client is self-registered, but the consent page presents only the attacker-chosen `client_name` (authorize.html:251) with no redirect-URI host, client_id, registration date, or 'unverified application' notice. `authorize_get` passes neither redirect_uri host nor any trust indicator to the template (routes.py:396-411).

**Impact**

An attacker registers a client named 'Claude' with an https redirect to their own host (only https/no-fragment is checked, routes.py:61-66) and sends a panel user a crafted /authorize link. The page is indistinguishable from the real connector's; one Approve grants the attacker's PKCE-bound code, which they exchange for read (or, if the user selects it, readwrite) access to that user's entire vault. The read-only preselect (#63) limits the default blast radius but not the read exfiltration.

**Recommended fix**

Render the redirect URI's host and the client_id on the consent card and add a standing notice that the application registered itself and is not verified by the server. Optionally maintain an operator-configurable allowlist of known connector redirect hosts (claude.ai, chatgpt.com) to show a 'known client' badge, and show a stronger warning for everything else.

**Independent verification**

Confirmed. /register is on the unauthenticated api router (docker-compose.yml:73; live GET /register returns 405 allow: POST from uvicorn with no forward-auth redirect), so any party can register a client with any https redirect (_valid_redirect_uri routes.py:61-66). authorize_get passes only client_name/client_id/redirect_uri as hidden form fields (routes.py:396-411) and src/control_panel/templates/authorize.html:250-260 renders only client_name and the requested level; grep of the template for unverified/trust/registered-itself finds nothing beyond the read-only preselect note. /authorize sits behind chain-oauth@file (docker-compose.yml:66; live probe 302 to Google), which restricts victims to legitimate Google-authorized panel users - exactly the population that holds vaults - so the phishing consent scenario is real. Mitigations that exist (read-only preselect #63, exact redirect match, PKCE) limit the default to read exfiltration, which in a multi-tenant vault of record is still a medium.

[...]

Issue #90, opened 2026-08-20
users: delete_user permits self-deactivation once another admin exists — contradicts the #69 self-edit lock


maxkuminov:
PR #80 blocks self-demotion/self-deactivation on the edit form (#69) and serializes admin-flag changes. `delete_user` (soft delete) still lets an admin deactivate their own account when another active admin exists. Decide whether the #69 promise is about the form or the account; if the latter, refuse self-delete in `delete_user` under the same advisory lock and actor re-check.

PR #223, opened 2026-09-05
openspec: propose panel-sessions-and-consent

maxkuminov:
OpenSpec proposal only — no application code, no tests. Closes nothing on merge; the three issues close when the change is implemented and archived.

## What it proposes

**#198 (V7.4.1) — a server-side session registry.** Today `logout()` calls `request.session.clear()` and nothing else, so the already-issued cookie stays a valid credential for seven days; the only server-side invalidator is `users.session_version`, bumped solely by the admin password reset. The proposal adds a `user_sessions` table (migration **024**, `down_revision = "023"`) holding one row per live session: the **SHA-256 of** a 256-bit CSPRNG identifier as the primary key, `user_id` with `ON DELETE CASCADE`, created/last-seen/expiry timestamps, `revoked_at`, and a forensic-only user-agent hash. The signed cookie carries the identifier; `get_active_session_user` resolves the row and refuses a missing, revoked or expired one on top of the existing checks; logout stamps `revoked_at`; the existing cleanup job purges on the OAuth retention rule. `session_version` is kept as a second, account-wide switch. The audit note's caveat is honoured: a `sessions_invalid_before` timestamp is explicitly rejected, with reasons.

**#197 (V6.2.2) — self-service password change.** `GET /admin/account` and `POST /admin/account/password` on the panel router (`require_user_panel`, not admin), CSRF-protected, verifying and hashing through the existing functions — the 72-byte truncation and the NUL rejection are preserved exactly, not "fixed". On success: new hash, `session_version += 1`, every session of that user revoked, and the current browser immediately re-issued a fresh one. Throttled at the login rate under a `(client IP, session user id)` key. One `MIN_PASSWORD_LENGTH = 12` constant shared with bootstrap and the admin reset, which are at 8 today; no composition rules, no forced rotation. The admin reset stays as recovery.

**#183 (V10.4.7) — the consent card identifies the client.** `authorize_get` also passes the redirect URI's **host** (`urlparse().hostname`, never `netloc`, so `https://claude.ai@evil.example/cb` reads `evil.example`; punycode, never Unicode-decoded), the `client_id`, and the registration date. An **unconditional** notice says the application registered itself and is not verified here. `OAUTH_KNOWN_REDIRECT_HOSTS` (default `claude.ai,chatgpt.com`) drives a badge worded about the *destination*; everything else gets the warning naming the host. Matching is **exact host equality** — no suffix, prefix, substring or wildcard — and the config validator rejects pattern-looking entries. No CSP; inline styles and `_theme.html` tokens as the house requires.

## Contents

`proposal.md`, `design.md` (18 decisions with rejected alternatives, the session lifecycle table, the password-change flow, the consent card contents, risks, non-goals, an Accepted-limitations table, and an Owner-decisions list), spec deltas for `panel-session-registry` (new), `panel-account-password` (new), `oauth-authorization-integrity`, `panel-user-administration` and `schema-integrity`, and `tasks.md` with six disjoint file-scoped slices in three waves plus a verification section.

`openspec validate --all --strict` is clean (29 passed, 0 failed).

## Coordination

- **Migration numbering:** 024 chains from **023**, reserved by a sibling change. It must not merge ahead of it; if 023 is abandoned, rebase onto 022. `HEAD_REVISION` in the schema gate moves to `"024"` only once both are in.
- **`security-event-logging` (PR #216):** event names are taken from that change's catalogue (`panel_logout` is already there; this adds `panel_session_replay_refused`, `panel_sessions_revoked`, `panel_password_changed`, `panel_password_change_refused`, using only fields already in its `ALLOWED_FIELDS`). If it lands first these emit through `security_events`; if not, the same names go through a plain module logger with the identifiers in the message text. No session identifier, password or hash is ever a field.
- **Deploy consequence, deliberate:** pre-registry cookies are refused rather than grandfathered, so the deploy signs every live panel session out exactly once. Grandfathering would keep the #198 window open for another seven days.

Refs #198 #197 #183


maxkuminov:
## Revised after Codex round 1 (FAIL — 3 BLOCKER, 11 MAJOR, 3 MINOR)

All seventeen findings folded; **none rejected**. Pushed as `1179c06`. `openspec validate --all --strict` clean (29/29). The finding→decision map is the last section of `design.md`.

### Blockers

- **Stale-read password change** — the handler verified and wrote against the `User` a FastAPI dependency had loaded, so an administrator's reset or deactivation committing in between was silently overwritten and access the admin had just removed came back. It now takes the **same** account-guard advisory key `_lock_admin_guard` uses, re-reads `FOR UPDATE` with **`execution_options(populate_existing=True)`** (without it SQLAlchemy hands back the identity-mapped object's pre-lock values and the re-read proves nothing), re-checks `is_active`, and verifies against the freshly read hash. The key moves to `src/oauth/grants.py` — already this codebase's home for advisory-lock primitives, and importable from both `routes.py` and `users.py`, which cannot import each other — with its **value unchanged**. New scenarios cover concurrent reset and concurrent deactivation, both asserting nothing is minted.
- **Uncommitted mint** — `get_session` neither commits nor rolls back, so an insert left to the caller could silently not happen and the cookie beside it would authenticate nothing. `start_session` now **owns its commit** and returns only when the row is durable; the revoke helpers still deliberately do not, because they must ride their caller's critical section. Per-caller placement is a table (login after its own commit; bootstrap **after** the bootstrap transaction, never inside it — that transaction holds `USER_BOOTSTRAP_LOCK_KEY`; the change in a second transaction). Requirement: the cookie authenticates the **very next** request.
- **Two pool leases per request** — folded exactly as the owner decided. The throttled `last_seen_at` UPDATE now runs on the **request's own** session, `GET`/`HEAD` only, committed inside the dependency before the handler runs, and skipped otherwise. Two extra reasons are written down: with `pool_size=5, max_overflow=10` a second lease halves an already-shared pool (`pool_timeout=30`, then 500s for every other caller), and committing in the dependency releases the row lock before any handler can request the account-guard lock — holding it across that wait deadlocks against an administrator revoking that same user's sessions. A concurrency test drives more concurrent stale-session requests than `pool_size + max_overflow` and asserts no pool timeouts.

### Majors

- Change-password throttle is now **two stacked limits** — one keyed on the account, one on the client address. The old composite key gave an attacker a fresh allowance per rotated address.
- `_valid_redirect_uri` / `register_client` brought into scope: **non-empty hostname + successful IDNA-to-ASCII conversion**, stored converted. `https://@/cb` (authority `"@"`, host empty) and a Unicode host are both tested. Pre-existing rows may still be hostless, so the display path renders "destination could not be determined" and always takes the warning branch.
- `create_user` joins the shared password policy — it accepts 8 chars today and passes input straight to `hash_password`, so a NUL is a latent 500. Four setters now, one validator.
- `User.sessions` specified as `cascade="all, delete", passive_deletes=True`, with the permanent-delete test driving the **real handler** and asserting the ORM issues no per-row delete.
- Purge moves to **max(`expires_at`, `revoked_at`)**. The OAuth `expires_at`-only argument does not transfer: an admin reset revokes already-expired rows too, and those would be purged on the next tick — the #64 blank space, reintroduced. Expired-then-revoked scenario added.
- Schema-gate head is now a **`## MODIFIED Requirements`** block on the existing requirement (heading kept byte-identical so it modifies rather than duplicates), moving `017 → 024` and stating plainly that 024 is ordered after `index-integrity-hardening`'s 023.
- Migration 024 rewritten to 016/017/022's shape: `COMMENT ON TABLE` marker mirrored in the model, exact reconciliation resolving the FK delete action through `pg_constraint.confdeltype` and indexes through `pg_index`, **refusal** of a foreign or partial shape, marker-guarded downgrade, and seeded rows preserved across a stamp-back re-run.
- New **slice 7 (integration)** owns `src/services/security_events.py` — `security_events` is now assumed to **exist** (sibling Slice A in flight), so the four new events are **declared in `EVENT_FIELDS`** rather than emitted past it — plus the thirteen existing test modules that mint or assert a browser session (`test_auth_routes_real_hasher.py`, `test_issue_4_authorize_post_redirect_validation.py`, `test_security_review_followups.py`, the `/authorize` suites, `tests/integration/test_oauth_grants_pg.py`, …), which get **one** shared mint helper rather than thirteen hand-built cookies.
- Validation now requires `row.user_id == cookie user_id`; a mismatch refuses and clears.
- Logout failure path: the rollback is itself guarded, and the record carries the exception's **class name only** — SQLAlchemy renders bound parameters, one of which is the stored session hash. Tested with a hash-bearing exception and with a failing rollback.
- Secret-absence widened to the identifier **and its stored SHA-256** (the registry key names a specific live session) and to any substring ≥ 12 characters, aligned with the logging change's canary rule.

### Minors

`SESSION_TOUCH_INTERVAL_SECONDS` and `SESSION_PURGE_RETAIN_DAYS` are `ge=1` with invalid-config scenarios; the "single-user mode never touches `user_sessions`" statement is now scoped to the authentication surface, exempting the maintenance purge; allow-list entries are stripped of outer whitespace and rejected for internal whitespace.

maxkuminov:
## Revised after Codex round 2 (FAIL — 1 BLOCKER, 1 MAJOR, 2 MINOR)

All four folded, none rejected. Pushed as `3f5e9e4`; `openspec validate --all --strict` clean (29/29). Final spec round.

**BLOCKER — the mint could create a session for an account just deactivated.** Round 1's fix made `start_session` commit, but it committed *after* the caller's guard had been released and with no fresh check: the login mint held no guard at all, and the post-change re-issue ran in a second unguarded transaction. So a deactivation landing during a paused login, or in the gap between a password change's commit and its re-issue, is followed by an insert creating a **live row for a disabled account**. Validation refuses that row while `is_active` is false — which is exactly what hides it — and on reactivation it becomes a working credential nobody granted. `start_session` is now one critical section: take the account guard, re-read the user `FOR UPDATE` with `populate_existing=True`, require the row to exist and be active, insert, commit (releasing the guard), and only then write `sid` into the cookie. The bootstrap key and the account guard are taken sequentially and never nested, so no cycle is introduced. Three scenarios added — deactivation racing a login, deactivation racing a post-change re-issue, and a **later reactivation** confirming no row from that window exists. Recorded as a limitation that every login now waits on the guard: the contention set grew, but the section is one indexed read plus one insert and production is two users deep.

**MAJOR — slice 7's inventory.** Added `tests/test_issue_69_self_edit_role_lock.py` and `tests/test_issue_90_self_delete_refused.py`, which break on the guard-key move and on `revoke_user_sessions` returning a row count (both drive the admin handlers against fake sessions whose shapes encode what those handlers execute — a fixture repair, not a rewrite, and they keep asserting the #69 and #90 guarantees). The shared mint helper is named explicitly: `tests/session_helpers.py`, created and owned by slice 7 so no other slice races it.

**MINORS.** `/admin/account` now 404s in single-user mode on **both** methods — a page whose only content is a form that cannot exist there is not a page, and the sidebar entry is already gated on `multi_user_mode`; decision, task and scenario aligned. And rate-limit rejections are explicitly **exempt** from the flash-and-303 rule: slowapi's `RateLimitExceeded` never reaches the handler and the application-wide JSON 429 stands, because the alternatives are a second in-handler counter that would diverge from the limiter that already decided, or replacing a process-wide error handler for one route. A sixth-request test asserts the limiter's own response and that nothing is written.

Issue #205, opened 2026-09-04
[ASVS V2.3.4] No caller-visible write precondition: an agent editing from a stale read_note silently overwrites concurrent changes (and the edit_note docstring says it is protected)


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 — Codex cross-family pass, independently verified with measurements before filing.

**Location:** `src/mcp_server/tools.py:2048-2057` (`edit_note_impl`), `:3673-3678` (`write_file` overwrite passes no `expected=`), `src/mcp_server/read_result.py:461-490` (`ReadNoteResult` has no hash), `src/mcp_server/server.py:448-452` (docstring)

**What is wrong**

No write tool accepts an expected hash / version / mtime, and `read_note` returns none (`notes_metadata.content_hash` is never surfaced). The existing `expected=` compare-and-swap is server-internal: `edit_note` reads the file inside the call and passes those bytes to `_atomic_write_at`, so it covers only the tool's own read→rename interval and structurally cannot see a change that landed between the agent's `read_note` and its `edit_note`. Blast radius depends on mode — `append` loses nothing, `find` refuses on zero/non-unique matches, `section` replaces only the resolved body, `set_frontmatter` merges — so the whole-file lost-update paths are full-replace / `replace_frontmatter` and `write_file(overwrite=True)`, the latter with no check at all even in-call (documented in `vault-tools.md:219`). Nothing documents lost-update from stale client state as accepted, and the MCP docstring over-promises: it tells the agent that "a note somebody else changed in the meantime fails with `File changed while editing`… re-read and retry", which an agent doing the recommended read → edit round trip will read as covering its own cycle.

**Impact**

Silent, unrecoverable loss of a concurrent edit (second key on the same root in multi-user mode, or the owner editing in Obsidian on the rw mount) during the agent's deliberation window — the product's #1 expensive failure class, though a missing affordance rather than a server bug.

**Recommended fix**

[...]

Issue #154, opened 2026-08-27
Non-finite YAML floats (NaN/Infinity) in frontmatter are Python-representable but not valid JSON; JSONB and the structured read_note view may reject them


maxkuminov:
Split out of #149's round-4 adversarial pass, deliberately left out of scope there.

The #149 scrub at the frontmatter parse boundary (`_scrub_frontmatter`) removes only values *nothing* can render: non-encodable strings, `str()`-raising integers, self-referential containers. `NaN`/`Infinity` floats are a different predicate — Python renders them fine (`str()`, `repr`, pydantic stringification), but they are not valid JSON, so:

- PostgreSQL JSONB (`notes_metadata.frontmatter`) will likely reject them at index time;
- `pydantic_core.to_json` emits non-RFC output for them, which could desynchronize `read_note`'s text block from `structuredContent` or break strict clients.

Pre-existing (not introduced by #149), unreproduced end-to-end so far. Scope for a proposal: decide the boundary rule (coerce to strings at parse like dates? drop with a `lossy` reason?), check what the indexer actually does today with `x: .nan`, and pin whichever answer through the indexer, the panel, and the structured read path.

Refs #149.

Issue #126, opened 2026-08-25
BLOCKER: an unsanitized frontmatter `title` freezes every index pass


maxkuminov:
Verified (adversarial pass, CONFIRMED BLOCKER). Found by #122 tier 1b.

`indexer.py:1071` assigns the frontmatter `title` straight to the row; only the separate JSONB `frontmatter` copy is sanitized. `notes_metadata.title` is `VARCHAR(512)` (`db.py:283`). A note with `title: 2026-08-25` (YAML parses to `datetime.date`), a list, or a >512-char title makes the batch INSERT (`indexer.py:1205`) raise, aborting the whole pass transaction. Because nothing commits, the content hash never advances, so **every subsequent 5-minute tick retries the same fatal batch forever** while search keeps answering from the last good rows — a silent, self-perpetuating freeze.

Fix: coerce and bound — `title = str(sanitized_value or stem)[:512]`. Small and safe; verified no invariant conflict.

Refs #122

Issue #149, opened 2026-08-26
read_note's response envelope is forgeable by note content; section body needs structural framing


maxkuminov:
Split out of #140, where it was found and reproduced during the pre-implementation spec audit. #140 deliberately ships **without** documenting how to recover a section's body from a `read_note` response, because every textual procedure proposed for it turned out to be forgeable. This issue is that problem on its own.

## The defect

`read_note` renders every response as an envelope — `# <title>`, `**Path:**`, optional `**Tags:**` and `**Frontmatter:**` — followed by `\n---\n` and the selected content (`src/mcp_server/tools.py`, the `parts` list). Every component of that envelope is **note-controlled and interpolated unsanitized**, so a valid note can emit a line that mimics the envelope's own separator.

Two confirmed reproductions, on different fields:

**1. A multiline YAML title**

```yaml
---
title: |-
  [Project](https://example.invalid)
  ---
  injected
---
# A
old
```

`read_note(path, section="#1")` renders a bare `---` line above the real separator. Splitting on the first `\n---\n` and dropping a line yields a remainder beginning ``**Path:** `n.md` `` — written back, it clobbers the section.

**2. A quoted frontmatter key** (found after the title was patched)

```yaml
---
"safe\n---\nforged": value
title: Safe
---
# A
old
```

The key renders across two lines inside the `**Frontmatter:**` block. The same extraction yields `\n---\n# A\nold\n`.

## Why the obvious fixes don't work

- **Sanitizing the named fields** doesn't close the class — two successive audit rounds each found a *different* field. Title, path, tags, frontmatter keys and values are all interpolated, and so is whatever gets added next.
- **A per-component invariant doesn't compose.** "No component contains a line terminator" does not imply "the envelope contains no separator": a component whose value is exactly `---` needs no newline at all, and forges one the moment any field is rendered alone on a line. Today's layout prefixes every field, so this is latent rather than live — but the invariant can't promise what it appears to.
- **Collapsing terminators to spaces is lossy**, and trades failure class 1 for failure class 2: the distinct paths `a\nb.md` and `a b.md` render identically, plausibly pointing an agent at the wrong note.

## The shape of the fix

Structural framing, so nothing needs parsing: return the section as its own field in a structured MCP result — `metadata` / `heading` / `section_body` — and let the caller pass `section_body` straight to `edit_note(section=…)`.

This changes `read_note`'s response shape and touches every consumer, which is why it isn't riding along on #140. Needs its own proposal.

[...]

PR #224, opened 2026-09-05
openspec: propose write-preconditions

maxkuminov:
OpenSpec proposal only — no application code, no tests. `openspec validate --all --strict` passes (29/29).

**#205 — no caller-visible write precondition.** The `expected=` compare in `_atomic_write_at` is server-internal: it covers *this call's* read → rename and structurally cannot see a change that landed between the agent's `read_note` and its `edit_note`. This proposes an **optional** `expected_hash` on every overwrite path — `edit_note` (all four modes, `dry_run` included), `set_frontmatter`, `write_file(overwrite=True)`, `move_note` (binding the source note only) and `delete_note` — checked immediately after the in-call read and before mode dispatch, the size cap, the diff and every no-op branch. A mismatch is an in-band refusal naming the file's **current** hash with nothing written; a precondition that cannot be enforced (a no-clobber creation, an incumbent over the read cap) is refused, never ignored. The existing in-call compare stays: two windows, documented as a pair.

**The digest, once:** `"sha256:" + hex(SHA-256(complete raw bytes of the file on disk))` — the same bytes `expected=` compares and the same digest `vault_fs.fingerprint` already computes. Deliberately **not** `notes_metadata.content_hash`, which hashes universal-newline-*translated* text and so differs for every CRLF note; the prefix makes an accidental comparison fail uniformly. `read_note` returns it as a `content_hash` field — always the whole file's, for section and truncated reads alike, computed from the same read that builds the response (never a second one). Raw files get it via `read_file`'s existing base64 header, a new `hash_only=True` metadata-only response, and every write's success line — never an envelope around a text read.

**#154 — non-finite YAML floats.** Today, from the code: the read view already coerces (`_view_leaf` → `str(value)`), so the structured path is safe but silently spells it Python's way; the **indexer does not** — the float reaches the `frontmatter` JSONB column through SQLAlchemy's default `json.dumps`, which emits `NaN`/`Infinity`, which PostgreSQL's `jsonb` parser rejects, aborting the pass's single transaction with no per-note retreat, so nothing commits, `content_hash` never advances, and every tick retries the same fatal batch (#126's failure mode, whole tenant). The rule: leave `_scrub_frontmatter` alone (YAML and Python both render these; a coerced string there would make `set_frontmatter` rewrite `x: .nan` as `x: '.nan'`), coerce at the two JSON boundaries to the **YAML** token (`.nan`, `.inf`, `-.inf`), disclose it in `metadata_omissions`, never fail a pass, and keep `set_frontmatter`'s byte-identical round trip — which PyYAML gives for free while the mapping still holds the float.

No migration, no new dependencies. `design.md` carries the per-tool table, the two-window ordering, the refusal shape, the #154 today-behaviour analysis, rejected alternatives, an accepted-limitations table (L1–L8) and seven owner decisions with defaults chosen. `tasks.md` splits into four file-scoped slices (A digest+read → B note writes, C raw files, D #154 in parallel) plus the gates, including the mandatory adversarial pass on the destructive path.

Refs #205 #154


maxkuminov:
Revised at `08e5665` after Codex's pre-code review (FAIL: 2 BLOCKER, 10 MAJOR, 4 MINOR). `openspec validate --all --strict` still passes (29/29). **All 16 findings folded; none rejected.**

**BLOCKERs**
- *Typed refusals.* Adopted the sibling `mcp-rate-limits` contract verbatim (`src/services/refusals.py`, final line-initial `MCP-REFUSAL {…}` sentinel appended to prose, carried into structured results through `refusal_result`) instead of the prose shape. Four codes contributed to its closed set: `stale_precondition` (+`current_hash`), `no_incumbent`, `malformed_precondition`, `precondition_required` — each with `path` and `nothing_written: true`, and **no retry delay on any of them** (per the sibling's rule that a number invites a loop that cannot end). Dependency stated: if `refusals.py` has not landed, Slice A creates it with the identical contract.
- *`create_note` could not refuse.* `expected_hash` is now on `create_note`'s signature in both layers, tracked, and answered with `no_incumbent` **before any filesystem work**. New requirement text makes the general rule explicit: a signature that rejects the argument answers with a protocol error instead of the contract, which is the opposite of the promise.

**MAJORs**
- Canonical-only input (`sha256:<64 lowercase hex>`); bare/uppercase hex is `malformed_precondition`. The "fails uniformly" claim is gone — the useful property is that handing over the index column's bare hex is *refused as malformed*, which says what actually happened.
- `_sanitize_value` splits into a JSONB sanitiser and a title coercion over one shared canonical-token helper; new `index-integrity` requirement fixes **one** title rule across indexer / `read_note` / panel, with cross-tool tests and a byte-identity corpus for ordinary titles.
- `write_bytes_at` / `write_bytes` gain `expected=None` forwarding to `_atomic_write_at` (they had no such parameter, so the raw-byte path could not reach the comparison at all) — assigned to Slice C with a single-shot regression test.
- Slices resequenced: shared helper moved into **A**, merge order **A → (B ∥ D) → C**, with C after B because both edit `tools.py`/`server.py`. Ownership restated at the top of `tasks.md`.
- `write_file(overwrite=True, expected_hash=…)` on a missing path → `no_incumbent`, nothing created (it must not fall through to creating).
- New `metadata_coercions` list on the read result for the retained-but-coerced case; `metadata_omissions` keeps its "dropped whole" contract. Delivered as a MODIFIED read-result requirement.
- `frontmatter_yaml` described as **authoritative but LF-normalized**, with `read_file` (base64) named as the byte-exact route.
- Canonical output tokens regardless of source spelling, with `.NaN` / `.INF` / `+.inf` scenarios in both the read and index specs.
- Per-tool RESULT column added: `move_note` reports the **destination's** hash (rewrite-enabled moves report only the moved note's); `delete_note` reports none; dry runs and no-ops report none.
- `WRITE_PRECONDITION_REQUIRED` (Settings boolean, **default false**) with the typed `precondition_required` refusal, `create_note` / no-clobber exempt, carrying `current_hash` only where the tool had already read the incumbent. Recorded as owner decision O8.

**MINORs** — L10 records that "nothing written" excludes the call's own `usage_logs` row; L11 limits the compatibility claim to mutation/conflict semantics (success strings gain `content_hash`, and the existing assertions on them get updated in task 2.6); `read_file` argument precedence is now normative (`encoding` validated first, then `hash_only` vs `offset`/`limit`, then range checks); the "hash advances" scenario now mutates the note between the two passes.

maxkuminov:
Round 2 folded at `eb753df`. `openspec validate --all --strict` still passes (29/29). **9 of 10 findings folded as given; one folded with a reasoned deviation on half of it (below).**

**BLOCKER — `delete_file`.** In scope, both modes. The first draft's "raw tools are transport" reasoning contradicted this change's own claim to cover every content-destroying path: `permanent=True` is irreversible and `permanent=False` hides the bytes under a generated `.trash` name. New `file-access` requirement with stale / matching / required scenarios in both modes; the comparison runs before the trash rename and before the unlink, through the anchored beneath-root lookup that tool already validates with (not `open_mutable`). Owner decision O10; Slice C owns it.

**MAJORs**
- (a) **`concurrent_write`** joins the closed code set for the in-call window. Prose (`File changed while editing: …`) is unchanged and the sentinel appended — the sibling's additive rule — so existing assertions hold while the two windows are distinguishable **by code**. `move_note`'s per-source partial-success report is explicitly not this refusal.
- (b) **`precondition_unavailable`** for an over-cap incumbent, carrying `cap_name` and `cap_bytes`. "I could not check" and "I checked and it differs" no longer share a code.
- (c) **One precedence ladder**, now its own requirement: syntax → no-incumbent → unavailable → required → comparison → publication. Malformed always wins, decided before any filesystem work. Four combination scenarios (malformed + create, + no-clobber, + missing path, + over-cap) plus one more the ladder forced out: required mode on an unhashable file answers `precondition_unavailable`, not `precondition_required`, because supplying a hash there is impossible.
- (d) **`move_note` result matrix** — plain move / rewrite published / **moved note's own rewrite failed** (report the rename's unrewritten bytes, never the intended ones) / backlink-source failure (unaffected). The two failure modes are tested separately.
- (e) **Title rule named: the indexer's `_note_title` as it stands** — falsy fallback, container stringification, the 512-character bound — with the single non-finite exception, adopted by `read_note` and the panel. The resulting non-NaN changes are listed with expected outputs in design D10b and pinned as scenarios: `[2026-08-25]` → `['2026-08-25']`, a non-string key → `{'1': 'a'}`, a 600-character title → first 512; top-level dates, string lists, numeric and all falsy titles unchanged. Owner decision O11, limitation L13.
- (f) **Keys go through the token helper too** (`_view_key` and the JSONB key path). Collision handling **splits by boundary**, and this is the one deviation: JSONB takes **first-key-wins** as instructed (today's dict comprehension silently keeps the *last*, so this is a real fix); the **read view keeps omit-whole-view** with a reason code distinguishing a coercion-induced collision from a native one. First-wins in the view would silently drop a key the note contains, which is exactly what #149's rule forbids — a partial view is indistinguishable from a complete one. The difference is that the index cannot report a loss and the view can. Recorded as owner decision O12 and limitation L14; say the word and I will flip the view to first-wins.
- (g) **The base64 header's path is JSON-encoded**, uniformly for every path, in both the base64 header and the `hash_only` result — a filename may contain a newline (the framing requirement insists `a\nb.md` stays distinguishable), so an unescaped path could inject the very `content_hash:` line this change asks callers to trust. Tests cover newline, CR, colon and blank-line paths. Uniform rather than conditional so a reader never has to work out which form it is looking at; that changes the header line for ordinary paths too (L12, O13).

**MINORs** — `hash_only` conflicts are decided **by value** (`offset != 0 or limit is not None`, since MCP arguments arrive with defaults applied), with a scenario asserting `offset=0` succeeds and `limit=1` is refused; the `metadata_coercions` reason code is the literal **`non_finite_float`**, named in the spec and asserted in tests.

maxkuminov:
Round 3 folded at `07c128b` — all four, none rejected. `openspec validate --all --strict` passes (29/29).

**Post-rename contract (MAJOR).** `move_note`'s result is now split by *observed cause* rather than by "the rewrite failed": a rewrite that failed without observing a change reports the rename's bytes, while one refused as `concurrent_write` reports **no hash at all** — the destination holds an editor's bytes this call never read, so the rename's hash is as wrong as the intended one — and the result instead states that the move completed, the rewrite did not, and the destination must be re-read. A post-rename failure is a partial success and may never carry `nothing_written: true` in any form, since the rename committed and a caller told otherwise goes looking for a note that has moved. Three post-rename cases are now tested separately (no-conflict failure, conflict failure, backlink-source failure); recorded as limitation L15.

**Bounded incumbent reads (MAJOR).** Every read this change adds goes through one shared bounded reader in Slice A: `fstat` the descriptor the tool already holds, compare against that tool's own cap (`MAX_NOTE_BYTES` for note tools, `MAX_FILE_READ_BYTES` for raw-file tools), then read through **that same descriptor** — so the bytes measured are the bytes hashed and no second pathname is resolved. Over cap is `precondition_unavailable` when a hash was supplied or required mode is on; **and when neither applies it is not a refusal at all** — the call proceeds exactly as today and reports no result hash, so nothing that works now stops working (stated as an explicit compatibility rule, plus the general "a hash is reported only when the bytes can be bounded", limitation L16). Over-cap tests added for every affected tool and mode in both slices.

**Per-code remedy prose (MAJOR).** One generic "re-read and retry" was wrong for three of the six codes, so each now states what resolves *it*: `no_incumbent` → call again without `expected_hash`; `malformed_precondition` → the canonical `sha256:<64 lowercase hex>` and where a valid one comes from; `precondition_unavailable` → the cap's name and value, that only an operator can raise it, and the transfer download route for byte-exact inspection; `stale_precondition` / `concurrent_write` / `precondition_required` → re-read and retry with the returned or a fresh hash. A scenario asserts the first three do **not** say "re-read and retry".

**Syntax check at the entry (MINOR).** Split into its own pure helper, `_precondition_syntax_error`, called at each guarded tool's entry ahead of path resolution, the leaf check and any read, so `malformed_precondition` wins over not-found, over a symlinked leaf and over the size cap — a caller told "not found" for an argument that was never valid fixes the wrong thing. Precedence tests for missing / symlink / over-cap added across all guarded tools.

Issue #182, opened 2026-09-04
[ASVS V10.4.5] Replayed refresh token does not revoke the grant family


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V10.4.5, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/oauth/routes.py:821`

**What is wrong**

`_handle_refresh` rotates correctly (new pair minted, old refresh row set revoked=True at :911), but when a refresh token that is already revoked is presented again the locking select at :811-819 finds nothing and the handler returns `invalid_grant` (:821-822). The grant_id was already resolved by the unfiltered lookup at :800-806 and the family lock is held (:809), yet `revoke_grant_family` is never called. No test covers refresh-token reuse detection (grep of tests/ for replay/reuse found none).

**Impact**

A stolen refresh token from a public MCP client (ChatGPT/claude.ai register with token_endpoint_auth_method 'none') can be exchanged by the attacker first; the legitimate client's next refresh then fails and it silently re-authorizes, while the attacker's rotated family stays live with the user's vault scope — up to readwrite — for as long as it keeps refreshing (30-day sliding window). The reuse signal that RFC 6819 / OAuth 2.1 use to detect exactly this theft is discarded.

**Recommended fix**

In `_handle_refresh`, when the first lookup returns a grant_id but the `revoked == False` select returns no row for that hash, call `revoke_grant_family(session, grant_id)` and commit before returning invalid_grant (the lock is already held, so this is race-safe). Add a regression test: present the same refresh token twice and assert the second call leaves no live token in the family.

**Independent verification**

[...]

PR #225, opened 2026-09-05
feat(logging): security-event-logging Slice A — foundation (#190)

maxkuminov:
Slice A of the `security-event-logging` change: the logging foundation that Slices B, C and D import. The change directory (`openspec/changes/security-event-logging/`) rides along, with section 1 of `tasks.md` ticked.

## What landed

**`src/logging_setup.py` (new)**
- `configure_logging(level=None, fmt=None)` — walks the root's handlers and removes and closes every one **except** `error_log.installed_handler()`, installs one `StreamHandler(sys.stderr)`, sets the root level, re-applies the `httpx`/`httpcore` WARNING floors. Idempotent, never raises. Explicitly *not* `basicConfig(force=True)`, which would close the health page's ring buffer; the exemption is written down and both call orders are tested.
- `StructuredFormatter` — one physical line per record, UTC ISO-8601 `Z` timestamp, `msg` bounded, `stack` only with `exc_info` and elided **head 4 KiB + tail 3 KiB** with a marker naming the dropped byte count, the exception type line and the traceback's final line guaranteed present (each appended line itself bounded so the guarantee cannot unbound the record).
- Three disjoint name spaces per D18: `FORMATTER_OWNED`, `EMITTER_CONTROL`, `ALLOWED_FIELDS` (32 fields, declared type + string bound). A value failing its type check is **dropped, never converted** — booleans are not integers here. `error_type` is dual and the exception wins.

**`src/services/security_events.py` (new)**
- `EVENT_FIELDS` — the whole catalogue, 47 events, one entry per row of the architecture note's table (a test asserts neither side has a row the other lacks).
- `acquire(event, subject=None, *, level=WARNING) -> Permit | None` charges the allowance **once**; `emit(permit, …)` consumes it with no second check. There is no `should_emit`.
- `redacted_token_tag`, `client_ip(request)` (never a header), `subject_for(...)`, `flush_suppression_summaries()`, plus `strict_fields()` / `suppression_disabled()` / `reset_state()` for tests.
- The suppressor: 10 per `(event, subject)` and 50 per subject per 60 s, every level accounted including INFO, ≤ 512 keys per map, summaries emitted lazily before the next acquire, **before evicting an entry with an outstanding count**, and at shutdown; lock-guarded; fails open.

**Migrations and wiring**
- `src/mcp_server/auth.py` — all ten `auth_failure` sites move onto `security_events.emit`; `_redacted_prefix` delegates to the one definition; `key_prefix` → `token_tag`; the OAuth branches' `oauth_tokens.id` moves out of `key_id` into `oauth_token_id`; both gain `client_ip` and `route`. The bare no-Bearer 401 stays unlogged.
- `src/services/quotas.py` / `src/services/vault.py` — the four caller-triggerable events migrate, `error` → `error_type`. `vault.py`'s post-publish flush note stays on the bare logger.
- `src/main.py` — `configure_logging()` replaces `basicConfig`, positioned after the import block; `_rate_limit_exceeded_handler` is wrapped in one local handler that emits `rate_limit_exceeded` and returns the delegate's response unchanged; the suppressor's flush is registered on every lifespan exit path, sandbox-mode early return included.
- `src/mcp_stdio.py` — `configure_logging(level=WARNING)` (stderr matters there: stdout is the protocol channel) plus an `atexit` flush.
- `src/config.py` — `LOG_LEVEL` (validated against the standard level names) and `LOG_FORMAT` (`json` | `text`, refused at startup otherwise).
- Docs: `docs/architecture/security-event-logging.md` (new) and the `CLAUDE.md` index row.

## Tests

`3122 passed, 439 skipped` before → **`3188 passed, 439 skipped`** after; 66 new tests across the four files named in tasks.md. `npx @fission-ai/openspec@1.3.1 validate --all --strict` — 27 passed, 0 failed.

Notable coverage: the AST sweep over every module under `src/` (rejecting unknown names, formatter-owned names, per-event violations and any dynamic field set); the real-process subprocess assertions for both `src.main` and `src.mcp_stdio`; the ring buffer in both call orders; eviction, shutdown flush and fail-open for the suppressor.

## Notes for the other slices

- **`src/oauth/routes.py` (Slice B) has two call sites the allow-list does not cover.** The refresh-reuse alarm from #182 (archived after this change was proposed, so it is not in the design's migration table) logs `event`, `error` and `revoked_tokens` through `extra=`. They sit behind a dated, per-name `_KNOWN_GAPS` entry in the sweep so the gap is visible rather than silent; Slice B should move both onto `security_events.emit` and delete the entry.
- Two files outside the Slice A list were touched, minimally and unavoidably: `tests/conftest.py` (`SETTINGS_ENV_KEYS` must list every `Settings` field or `test_conftest_env_hermeticity` fails) and `tests/test_issue_76_key_liveness_surfaces.py` (it greps `auth.py`'s source for `"reason": "inactive_user"`, which is now a positional argument; the ordering it asserts is unchanged).

Refs #190


maxkuminov:
Closing and reopening to retrigger CI (no workflow run was created for the head commit).

PR #268, opened 2026-09-06
feat(tools): write-preconditions Slice B — note write preconditions

maxkuminov:
Slice B of the `write-preconditions` change (OpenSpec). Slice A (#260) defined
the digest, the six typed refusal codes, the precedence ladder and the read
surface, and guarded **nothing**. This is where the ladder reaches the
destructive path.

## What landed, per tool

All five note write tools take `expected_hash: str | None = None`. Every
refusal writes nothing — asserted on disk, not promised.

| tool | binds | ladder placement | success reports |
| --- | --- | --- | --- |
| `edit_note` (all four modes, `replace_frontmatter`, `dry_run`) | whole file | syntax at entry; comparison immediately after the incumbent read, **before** mode dispatch, the size cap, the diff and every no-op | the bytes published; `dry_run` none |
| `set_frontmatter` | whole file | syntax at entry; comparison after `raw_bytes`, **ahead of** the defect diagnosis, the lossy refusal and the empty-`updates` no-op | the bytes published; a no-op none |
| `create_note` | — | syntax at entry, then `no_incumbent` **before any filesystem work**; exempt from required mode | the bytes published |
| `delete_note` (both `permanent` values) | whole file | syntax at entry; comparison before the pin, the `.trash` rename and the unlink | none — nothing remains to hash |
| `move_note` | **`from_path`'s bytes only** | syntax at the tool's entry, ahead of the descriptor gate; comparison in the preflight, before the rename, before any rewrite and before the index queries | the destination's, per the matrix below |

`move_note`'s refusal prose additionally states that backlink sources are *not*
bound — a precondition covering one of N files while implying all of them is
worse than none.

## The two windows, told apart by code

The existing in-call `expected=` compare is unchanged and now typed: its prose
is **byte-identical** (`File changed while editing: <name>`) with the
`concurrent_write` sentinel appended, so every existing `in`/`startswith`
assertion still holds while an agent can distinguish "you were stale on
arrival" (`stale_precondition`, carrying `current_hash`) from "the file moved
while I was working" (`concurrent_write`). A matching `expected_hash` does not
disable the second compare.

## `move_note`'s post-rename contract

| case | reported `content_hash` |
| --- | --- |
| plain move | the moved bytes |
| the moved note's own rewrite published | the post-rewrite bytes at `to_path` |
| that rewrite failed **without observing a change** | the rename's bytes — what is actually on disk |
| that rewrite **lost the in-call conflict** | **none at all**; the result states the move completed, the rewrite did not, and `to_path` must be re-read, rendered through `_concurrent_write_refusal(nothing_written=None)` so nothing claims nothing was written |
| a backlink source failed | unaffected — the moved note's |

The moved note is identified by `write_target is dst_target`, not by a path
comparison. An over-cap destination reports success with the hash omitted and
the reason named, rather than failing a call that already succeeded.

## Compatibility

`WRITE_PRECONDITION_REQUIRED` is wired through the four enforceable tools;
`create_note` is exempt. Under required mode `move_note`/`delete_note` do
perform the incumbent read, because that read is the only way to rank an
over-cap file above a missing hash — a test pins
`precondition_unavailable` beating `precondition_required` there.

An unguarded call is exactly today's call: the silent overwrite still happens,
`move_note`/`delete_note` read no incumbent (asserted by counting
`_read_incumbent`), and an over-cap file behaves as it does today (a move
succeeds and reports no hash; `edit_note` keeps its `Failed to read` message).
Only result **text** changed — every publishing write now appends its
`content_hash` (design L11).

## Docs

`docs/architecture/vault-tools.md` gains a "Write preconditions" section — the
digest defined once (pointing at the spec), the two windows as a pair, the
precedence ladder with the reason each rung is where it is, the bounded-
incumbent rule, the per-tool table with the result column, the post-rename
matrix, the six codes with their per-code remedies, and residuals
L1/L3/L6/L10/L11/L15/L16 — replacing the "no write tool accepts an expected
hash" paragraph #205 tracked. It also folds in slice A's two handoffs: the
`content_hash` row as a fixed 71-character allocation beside `path` in the
per-field budget table, and the note that the metadata budget's worst case is
now reachable via `tags` rather than `title`.

## Counts and gates

- `tests/test_write_preconditions_notes.py`: **103 tests** — every mode of
  `edit_note` × stale/matching/malformed; the four ordering cases (stale beats
  the diff, the no-op, the defect report and the size cap); the section write
  bound to the whole file; the two windows by code; `create_note`'s
  `no_incumbent` and malformed-beats-no-incumbent; both delete modes with
  `.trash` asserted absent on every refusal; the four `move_note` cases
  including the three post-rename outcomes separately; over-cap guarded vs.
  unguarded on all four reading tools; malformed beating not-found and a
  symlinked leaf on all five; required mode on each enforceable tool and
  `create_note` through it; `usage_logs` recording `expected_hash` and the
  refusal row.
- Three existing assertions updated (two `create_note` success strings, one
  `usage_logs` parameter allow-list) — the compatibility claim is about
  mutation semantics, not result text.
- `OMCP_ALLOW_SKIP_TRANSFER_INTEGRATION=1 pytest tests -W error::RuntimeWarning`:
  **4,439 passed, 542 skipped**.
- `openspec validate --all --strict`: 32 passed, 0 failed.

## Deliberately not here

`write_file` / `read_file` / `delete_file`, `write_bytes*`'s `expected=`, the
base64 header and `hash_only` (slice C); the `_sanitize_value` split and the
`_view_leaf` coercion (slice D).

Refs #205


maxkuminov:
Incorporated into merged PR #269. The remote commit 0d65dc2 is patch-equivalent to recovered commit 42d3723 (git range-diff reports equality); no unique work is omitted. PR #269 also includes subsequent review fixes, completed validation and production deployment.

Issue #196, opened 2026-09-04
[ASVS V4.1.2] Global HTTP-to-HTTPS redirect covers machine-facing API endpoints


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V4.1.2, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `docker-compose.yml:72`

**What is wrong**

Port 80 transparently redirects every path to HTTPS: `curl -sI http://<host>/mcp` returns `302 Found` with `Location: https://<host>/mcp` (same for /health, /token, /transfer/*). The redirect comes from a catch-all router defined on the Traefik container's own labels (`traefik.http.routers.http-catchall.rule=HostRegexp(.+)`, entrypoint `http`, middleware `redirect-to-https`), not from anything in this repo; the app's api-rtr router (docker-compose.yml:72-76) binds only to the `https` entrypoint and would otherwise not answer on port 80 at all. ASVS asks that only browser-facing endpoints (here /admin, /authorize) redirect and that API endpoints refuse plaintext so an erroneous HTTP client fails visibly.

**Impact**

An agent or MCP client misconfigured with an http:// server URL sends its `Authorization: Bearer omcp_...` API key or OAuth access token in cleartext on the first request; the 302 makes the mistake look like a routing quirk rather than a hard failure. Because 302 converts POST to GET and most MCP clients do not follow POST redirects, the client will not silently succeed, but the long-lived credential has already crossed the wire unencrypted and may be captured or logged. HSTS (src/main.py:342) protects browsers only.

**Recommended fix**

On the Traefik side, exclude the API host or the machine paths from the http-catchall (e.g. a higher-priority `http` router for Host(<host>) && (PathPrefix(/mcp) || PathPrefix(/transfer) || Path(/token) || Path(/register) || Path(/revoke) || PathPrefix(/.well-known)) that returns a 4xx via an errors/plain middleware, or simply do not route the API host on the `http` entrypoint and let it refuse), keeping the redirect for /admin and /authorize only. Record the decision in DEPLOYMENT.md so other proxy setups replicate it.

**Independent verification**

[...]

Issue #189, opened 2026-09-04
[ASVS V15.3.4] X-Forwarded-For trusted from the entire <internal-ip>/16 including a shared Docker network; Dockerfile allow-list does not match the deployment


maxkuminov:
**Source:** OWASP ASVS 5.0.0 assessment, 2026-09-04 (requirement V15.3.4, severity **medium**). Full report lives in the vault, not in this repo.

**Location:** `src/main.py:303`

**What is wrong**

ProxyHeadersMiddleware trusts 127.0.0.1, 10/8, 172.16/12 and 192.168/16. The actual proxy network is <internal-ip>/24 shared with ~50 other containers, each able to connect to obsidian-mcp:8000 directly (no ports published, same network) and to Traefik (whose `forwardedHeaders.trustedIPs=<internal-ip>/24` trusts them too). Any such peer can set an arbitrary X-Forwarded-For that becomes request.client.host, the key for every slowapi limiter (src/limiter.py:4). Separately, Dockerfile:29 passes `--forwarded-allow-ips 172.16.0.0/12,10.0.0.0/8`, which excludes the real proxy subnet, so the uvicorn-level control is inert and the deployment works only because of the broader app-level list.

**Impact**

Internet clients cannot spoof their IP (verified: Traefik discards untrusted XFF), but a compromised co-tenant container (n8n/open-webui run user-supplied code by design) can rotate forged IPs to defeat the 5/min login limiter and brute-force panel passwords that guard other users' vaults, and can spoof IPs in usage/OAuth logs.

**Recommended fix**

Narrow trusted_hosts to Traefik's fixed address (assign it a static IP on a dedicated network) and remove the <internal-ip>/16 entry; make the Dockerfile `--forwarded-allow-ips` value match (or drop the CLI flag and keep one source of truth); consider keying the login limiter on username in addition to IP.

**Independent verification**

[...]

PR #273, opened 2026-09-21
docs: archive the seven deployed 2026-09-06 sweep changes

maxkuminov:
Bookkeeping only. Nothing under `src/`, `tests/`, `alembic/`, the Dockerfile or the compose files is touched — the diff is `docs/` and `openspec/` alone.

## What was archived

The seven OpenSpec changes implemented, merged and deployed to production on 2026-09-06 were archived sequentially, each followed by `openspec validate --all --strict` and a diff of the touched `openspec/specs/**/spec.md`:

1. `reject-mask-decided-links` → `2026-09-21-reject-mask-decided-links`
2. `security-event-logging` → `2026-09-21-security-event-logging`
3. `vault-root-overlap-guard` → `2026-09-21-vault-root-overlap-guard`
4. `write-preconditions` → `2026-09-21-write-preconditions`
5. `mcp-rate-limits` → `2026-09-21-mcp-rate-limits`
6. `index-integrity-hardening` → `2026-09-21-index-integrity-hardening`
7. `panel-sessions-and-consent` → `2026-09-21-panel-sessions-and-consent`

`openspec list` now reports no active changes; final `openspec validate --all --strict` is **30 passed, 0 failed**.

**No delta header or requirement name needed fixing.** No requirement or scenario that existed before an archive was lost, and no near-duplicate requirement was added. The harmonization the rollout record promised held: the three `panel-performance-views` "Per-tool latency aggregates" MODIFIED blocks and the two `schema-integrity` "The schema gate covers both migrations of this wave before deploy" MODIFIED blocks were byte-identical, so the later archives applied them as no-ops. `panel-sessions-and-consent` task 8.6 is confirmed — the `schema-integrity` MODIFIED block landed on the existing requirement in place, and the spec holds exactly one requirement of that name. The one superseded scenario is `schema-integrity`'s "Head at 017", deliberately replaced by "The gate asserts the current head and 023 is in the chain", "Head at 024" and "The earlier waves' cases still run".

## Stale checkboxes reconciled

Tasks were ticked only where the rollout record (`docs/issue-sweep-2026-09-06.md`), `git log` or the issue tracker evidences them, each with a parenthetical citing the evidence.

## NOT EVIDENCED at archive

Left unchecked and annotated in place with what is missing — a gate the record does not show is not claimed:

| Change | Task | Missing |
| --- | --- | --- |
| reject-mask-decided-links | 7 | version-3 re-derivation confirmation and a live graph-tool exercise; the fix merged as PR #270 after the release image was built |
[...]

PR #274, opened 2026-09-21
spec: asvs-edge-hardening proposal (#196, #189, #194 residual)

maxkuminov:
**Proposal only — no implementation code.** `openspec validate asvs-edge-hardening --strict` passes; all 8 active changes still validate.

Three still-open findings from the 2026-09-04 ASVS assessment, all at the **edge** — the layer between a client and the first line of application code. They are disjoint in files and in mechanism, so they ship as one change with three slices, one independent subagent each, rather than as three changes.

### Slice A — #196, machine paths refuse plaintext instead of redirecting

Investigated the live proxy read-only. Traefik is **v3** (the rule syntax and the available middlewares are what matter, and both are v3-wide); entrypoints are `http`, `https` and a separate internal API port; the catch-all is on Traefik's *own* container labels (`HostRegexp(`.+`)`, middleware `redirect-to-https`) with **no explicit priority**, so its effective priority is its rule length, **16**.

A `priority=200` router on the `http` entrypoint, in this repo's compose labels, carries an **`ipAllowList` middleware with `sourcerange=192.0.2.0/32`** (RFC 5737 TEST-NET-1) — the one built-in Traefik v3 middleware that terminates the request itself, answering **403 with no `Location`** and never contacting the service. `/admin` and `/authorize` keep redirecting. `/.well-known/acme-challenge/` is carved out with a negated matcher because an ACME HTTP-01 certificate resolver answers on that entrypoint. `Caddyfile.example` gets the same rule natively so the published reference is not weaker than production.

Rejected: an `errors` middleware (only fires on a *backend* status, so it forwards first), a tiny app route answering 4xx (carries the credential into the app over plaintext and spends app resources on unauthenticated traffic), editing the host's `http-catchall` (outside the tree, not reproducible from the repo — the same reasoning already recorded there about Traefik rate limits), and a Traefik plugin (needs host static config).

### Slice B — #189, one proxy-trust control plus a username-keyed login limit

`TRUSTED_PROXY_IPS` (pydantic, `NoDecode`, CSV-or-JSON, every entry through `ipaddress.ip_network`, malformed entries refuse startup) defaults to exactly today's effective list, so a deploy with an untouched `.env` changes nothing. The Dockerfile's inert `--forwarded-allow-ips` becomes `--no-proxy-headers` — *aligning* the two lists was rejected because uvicorn's `proxy_headers` defaults to enabled and reads `$FORWARDED_ALLOW_IPS`, so "two controls that agree" is breakable from the environment.

The username limit **cannot be a slowapi decorator**: `src/limiter.py` documents that `key_func` is synchronous, runs before the handler, and reading the form body there consumes the stream. So it is an in-process fixed-window budget in `src/services/rate_limits.py` reusing the salted fixed-size slot table from the `/mcp` failed-auth budget.

**Lockout-DoS answer** (the check runs *before* password comparison, which is what makes a lockout possible in principle): only failures count; the window is 15 minutes with nothing durable and no admin unlock; an already-authenticated session is untouched, so a flood cannot evict anyone; the threshold (10/15 min) is far above typo rates; and the refusal is **byte-identical** to an ordinary failed login — same 401, same body, not a 429 — so it is not a username oracle. What remains is stated, not hidden: a sustained flood can deny *fresh password sign-in* for one named account for the flood's duration plus the window.

Also fixes a pre-existing defect found while designing: `login_submit` takes an unbounded `username` and `panel_login_failed` writes it straight into the security log today, so an attacker can drive arbitrary bytes into it. Bounded to 255 (the `users.username` width).

### Slice C — #194 residual, expiry of unused dynamic OAuth clients

**A migration is needed (025, `oauth_clients.last_used_at`).** All three no-migration signals were considered and rejected on correctness:

- `user_id IS NULL` — `src/oauth/routes.py:819` says plainly that in **single-user mode** the client is never claimed, so every client in the deployment `DEPLOYMENT.md` walks operators through would look unused.
- Absence of child rows — a *used* `OAuthCode` is deleted immediately with no age gate, and tokens seven days past expiry, so "used and aged out" is indistinguishable from "never used". Kept as a mandatory *additional* guard.
- `usage_logs.actor_ref` — records tool calls, not token issuance, and silently couples OAuth retention to an analytics table whose retention may change later.

The **race** is real, not theoretical: a single `DELETE … WHERE NOT EXISTS` is unsafe because inserting a code takes `FOR KEY SHARE` on the parent, but when the `DELETE` unblocks under READ COMMITTED PostgreSQL re-evaluates only the target row's own predicate, not the subquery — so the delete proceeds and `ON DELETE CASCADE` removes the code just issued. The sweep therefore uses `SELECT … FOR UPDATE SKIP LOCKED`, re-checks inside the lock, and the authorization that loses returns a clean `invalid_client`, never a 500.

### Capabilities

New: `edge-transport-security` (5 requirements). Modified: `request-trust` (Restricted proxy header trust), `api-auth-hardening` (Login rate limiting), `security-event-logging` (+2), `oauth-authorization-integrity` (+2), `schema-integrity` (+3).

`design.md` carries an **Accepted limitations** section (9 items) so they are not re-fixed when re-reported. `tasks.md` follows the project pipeline: Codex spec review before code, three disjoint slices on `aeh-a-plaintext` / `aeh-b-proxy-trust` / `aeh-c-client-expiry` with the two shared-file seams called out, `openspec-verifier`, two adversarial Codex rounds with triage, `make deploy` with the compose file copied to the deploy dir, a `curl` acceptance matrix plus named MCP tool calls in place of a browser pass, and archive.

### Open questions for the owner

- Does anything poll `http://<host>/health`? Answerable only on the host; task 1.1 checks before deploy.
- Does #194 itself close? Its remaining item is the operator chore of setting daily limits on five grandfathered keys.
- 30 days for the client expiry, or shorter? 7 is also defensible.

Not for merge yet — Codex spec review (task 0.2) comes first.


maxkuminov:
Superseded by #275, which carries this proposal (rebased onto main after #273) plus its three implementation slices and the review fixes.

